PR #1, opened 2020-04-24
Allow gradients in fmap

DhairyaLGandhi:


willtebbutt:
Any chance of some documentation alongside this PR explaining what the purpose of the two-argument version of `fmap` / `fmap1` is about.

Additionally, why do we need a separate `functor_tuple` function, as opposed to just implementing a method of `functor` for `Tuple`?

DhairyaLGandhi:
Its to make Optimisers.jl work by returning a slightly trained version of the model. The tuple function is so that we return a tuple of layers in a chain. I should revisit to see if its necessary still but otherwise this should be merged so we can release Optimisers.jl

willtebbutt:
I agree that this should probably go in, but a comment of some kind explaining what you've said above would be great.

src/functor.jl
```diff
@@ -36,3 +54,8 @@ function fmap(f, x; cache = IdDict())
   haskey(cache, x) && return cache[x]
   cache[x] = isleaf(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache), x)
 end
+
```
DhairyaLGandhi, line 57:
```suggestion
# Allow gradients and other constructs that match the structure of the functor
# to allow for `map` style computations and return a modified version of the struct.
# This way we can use `fmap` to update the params with their gradients
```

DhairyaLGandhi: cc @willtebbutt

DhairyaLGandhi:
bump!

src/functor.jl
willtebbutt: LGTM!

willtebbutt (commented):
The changes look get. Perhaps some tests and a patch bump?

ToucheSir (approved):

ToucheSir:
Woops, meant to say LGTM once the tests land.

DhairyaLGandhi:
Allows writing something like

```julia
Functors.fmap(x, y) do x, y
  isnothing(x) && return y
  isnothing(y) && return x
  f(x,y)
end
```

where x and y are basically NamedTuples out of Zygote, or a mix of the actual object tree (think Chain, model struct etc) and the corresponding gradients.

src/functor.jl
```diff
@@ -44,10 +44,29 @@ Equivalent to `functor(x)[1]`.
 """
 children(x) = functor(x)[1]
 
+function functor_tuple(f, x::Tuple, dx::Tuple)
```
darsnack, line 47: I don't think `x` is a tuple
```suggestion
function functor_tuple(f, x, dx::Tuple)
```

src/functor.jl
```diff
@@ -44,10 +44,29 @@ Equivalent to `functor(x)[1]`.
...
+end
+functor_tuple(f, x, dx) = f(x, dx)
+functor_tuple(f, x, ::Nothing) = x
+
+# @functor Chain
+# Chain -> func = (layers = (Dense,Dense),), gs -> (layers...)
+function _default_walk(f, x, dx)
+  func, re = functor(x)
+  map(func, dx) do x, x̄
+    # functor_tuple(f, x, x̄)
```
darsnack, line 60: This should work with the other change?

darsnack (commented):
Looks good!

ToucheSir:
@DhairyaLGandhi this already has my approval, so feel free to merge any time :)

Issue #6, opened 2020-12-04
Passing field names to @functor for typed struct gives constructor error


andrschl:
I am not sure whether this is intended or not, but when trying to specify the trainable parameters in the following code a "no method matching error" for the constructor is returned. 
```
struct MyLayer{R, S, T}
    a::R
    b::S
    c::T
end
Flux.@functor MyLayer (a,b)
m = MyLayer(Dense(1,10,tanh), Dense(1,10,tanh), zeros(5))
Flux.destructure(m) # returns no method matching error for the constructor
```
In the last sentence of the corresponding section in the [documentation](https://fluxml.ai/Flux.jl/stable/models/advanced/) it is mentioned that a corresponding constructor has to exist. However, it is not clear to me how this constructor should look like.

devmotion:
This issue seems to be the same as the one discussed in https://github.com/FluxML/Functors.jl/issues/3. You can check the comments therein for a solution to your problem: Instead of `Flux.@functor MyLayer (a,b)` you can define
```julia
function Functors.functor(::Type{<:MyLayer}, x)
    function reconstruct_MyLayer(xs)
        return MyLayer(xs.a, xs.b, x.c)
    end
    return (a = x.a, b = x.b), reconstruct_MyLayer
end
```

CarloLucibello:
addressed in #7

Issue #3, opened 2020-05-07
Correct way to reconstruct an instance of an object


willtebbutt:
The following (unsurprisingly) yields an error:
```julia
struct Foo
    a::Float64
    b::String
end

@functor Foo (a,)

x, re = functor(Foo(5.0, "hi"))

julia> re(x)
ERROR: MethodError: no method matching Foo(::Float64)
Closest candidates are:
  Foo(::Float64, ::String) at REPL[34]:2
  Foo(::Any, ::Any) at REPL[34]:2
Stacktrace:
 [1] (::var"#22#23")(::NamedTuple{(:a,),Tuple{Float64}}) at /Users/willtebbutt/.julia/dev/Functors/src/functor.jl:12
 [2] top-level scope at REPL[37]:1
```
You can imagine this kind of thing cropping up, where only a subset of the fields of an object are considered "parameters", but you _do_ need to have access to the others to make an instance of the object.

Is the intended way to handle this to define my own method of `functor` as follows:
```julia
function Functors.functor(x::Foo)
    function reconstruct_Foo(xs)
        return Foo(xs.a, x.b)
    end
    return (a = x.a,), reconstruct_Foo
end

julia> x, re = functor(Foo(5.0, "hi"))

julia> re(x)
Foo(5.0, "hi")

julia> fmap(Float32, Foo(5.0, "hi"))
Foo{Float32}(5.0f0, "hi")
```
?

MikeInnes:
There are a couple of ways to handle this situation. Overloading `functor` is totally valid (although you probably want to overload `functor(::Type{<:Foo}, x)` rather than `functor(x::Foo)`; the former method is used under the hood so that we can handle both a `Foo` and its gradient the same way).

There are a couple of other ways. First, you can just ignore `String`s when you map over parameters, assuming only numbers and numerical arrays are relevant to what you're doing. e.g. `fmap(x -> x isa Array{<:Number} ? f(x) : x, model)`.

Another option is to create a wrapper around `functor`, e.g. in Flux [trainable](https://github.com/FluxML/Flux.jl/blob/36d3a9ce998a23a53c0fd93123874b52bd0a3f02/src/functor.jl#L40) falls back to the `functor` definition but can also be easily overloaded by layers. Then `params` walks over the output of `trainable`, not `functor`.

Hope that helps; unfortunately it's a bit use-case-dependent but I'd be happy to look at what you're doing in more detail if that would be useful.

willtebbutt:
[...]

PR #7, opened 2020-12-20
Reconstruct with all fieldnames

ToucheSir:
This ensures that important, non-fmappable properties with non-default values are preserved. Should address https://github.com/FluxML/Functors.jl/issues/6 and https://github.com/FluxML/Functors.jl/issues/3#issuecomment-676736669.

I believe this is theoretically breaking, but a [search](https://juliahub.com/ui/CodeSearch?q=functor&u=all&t=macro) on JuliaHub doesn't turn up any uses of the macro that specify a list of fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,25 @@ Foo(1.0, [1.0, 2.0, 3.0])
 
 `functor` returns the parts of the object that can be inspected, as well as a `re` function that takes those values and restructures them back into an object of the original type.
 
-For a discussion regarding implementing functors for which only a subset of the fields are "seen" by `functor`, see [here](https://github.com/FluxML/Functors.jl/issues/3#issuecomment-626747663).
+To include only certain fields, pass a tuple of field names to `@functor`:
+
+```julia
+julia> struct Baz
+         x
+         y
+       end
+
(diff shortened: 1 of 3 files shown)


devmotion:
I guess the sentence "For a discussion regarding implementing functors for which only a subset of the fields are "seen" by functor, see here." could be removed from the README and replaced with an example.

DhairyaLGandhi:
Seems sensible at a first glance. Thanks!

ToucheSir:
Updated docs per @devmotion's comment. @DhairyaLGandhi can we cut a release with this so that Flux can start using it? I've seen a fair few PRs lately that stand to benefit.

CarloLucibello:
@DhairyaLGandhi bump

DhairyaLGandhi:
Seems alright. This would make a breaking release though.

CarloLucibello:
tag?

ToucheSir:
Yup, I was thinking a 0.2 so as to make it in time for Flux 0.12 (or barring that, 0.13)

DhairyaLGandhi:
Cool, good that we're in agreement then

PR #10, opened 2021-02-22
Add predicate keyword to fmap

darsnack:
This adds a `predicate` keyword to control the descent of `fmap`. Now, `fmap(f, x; predicate = ...)` applies `f(x)` if `isleaf(x) || predicate(x)`, otherwise it uses `functor` to descend further into `x`. By default, `predicate` is has no effect on the previous behavior.

This addition lets us deal with issues like https://github.com/FluxML/Flux.jl/issues/1517 without needing to special case every possible leaf situation (which we might not want to do for certain types).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,3 +71,22 @@ Any field not in the list will not be returned by `functor` and passed through a
 It is also possible to implement `functor` by hand when greater flexibility is required. See [here](https://github.com/FluxML/Functors.jl/issues/3) for an example.
 
 For a discussion regarding the need for a `cache` in the implementation of `fmap`, see [here](https://github.com/FluxML/Functors.jl/issues/2).
+
+To control the depth of descent by `fmap`, the keyword `predicate` can be used:
+```julia
+julia> using CUDA
+
+julia> x = ['a', 'b', 'c'];
+
+julia> fmap(cu, x)
+3-element Array{Char,1}:
+ 'a': ASCII/Unicode U+0061 (category Ll: Letter, lowercase)
+ 'b': ASCII/Unicode U+0062 (category Ll: Letter, lowercase)
+ 'c': ASCII/Unicode U+0063 (category Ll: Letter, lowercase)
+
+julia> fmap(cu, x; predicate = CUDA.isbits)
+3-element CuArray{Char,1}:
+ 'a': ASCII/Unicode U+0061 (category Ll: Letter, lowercase)
+ 'b': ASCII/Unicode U+0062 (category Ll: Letter, lowercase)
+ 'c': ASCII/Unicode U+0063 (category Ll: Letter, lowercase)
+```
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ It is also possible to implement `functor` by hand when greater flexibility is r
 
 For a discussion regarding the need for a `cache` in the implementation of `fmap`, see [here](https://github.com/FluxML/Functors.jl/issues/2).
 
-To control the depth of descent by `fmap`, the keyword `predicate` can be used:
+Use `predicate` for more fine-grained control over whether `fmap` descends into a particular value:
 ```julia
 julia> using CUDA
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ It is also possible to implement `functor` by hand when greater flexibility is r
 
 For a discussion regarding the need for a `cache` in the implementation of `fmap`, see [here](https://github.com/FluxML/Functors.jl/issues/2).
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -71,3 +71,22 @@ Any field not in the list will not be returned by `functor` and passed through a
 It is also possible to implement `functor` by hand when greater flexibility is required. See [here](https://github.com/FluxML/Functors.jl/issues/3) for an example.
 
 For a discussion regarding the need for a `cache` in the implementation of `fmap`, see [here](https://github.com/FluxML/Functors.jl/issues/2).
+
+To control the depth of descent by `fmap`, the keyword `predicate` can be used:
```
ToucheSir, line 75:
```suggestion
Use `predicate` for more fine-grained control over whether `fmap` descends into a particular value:
```

ToucheSir:
Tried to think of a shorter word than `predicate` and came up empty :laughing:. Generally looks pretty sensible to me, just needs a short test.

darsnack:
Yeah I also thought maybe `descend(x)` but it seemed weird cause we actually want to write the check for "don't descend."

CarloLucibello:
Should we do 
```julia
cache[x] = predicate(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache), x)
```
and default predicate to `isleaf` for more granular control?

Or maye `predicate(x) && isleaf(&)` makes more sense?

Just asking, didn't think carefully about the implications yet

darsnack:
No I think we want `predicate(x) || isleaf(x)` because the default set of leaf nodes is what most people want. Usually you want to add more on top of what's already considered a leaf. Using the OR makes this more ergonomic, so people don't have to remember to always write their rule OR `isleaf(x)`.

CarloLucibello:
I have nothing else than the `gpu` use case in mind, so just playing the devil's advocate. If we go with this design, we have no way to exclude leafs and if the need arises in the future we will have to do an annoying breaking change. Anyways, if you are (moderately) confident, I'm ok also with OR

darsnack:
Fair point. I'll change it to the default style unless @ToucheSir had an opinion.

src/functor.jl
```diff
@@ -38,7 +38,7 @@ end
 
 # See https://github.com/FluxML/Functors.jl/issues/2 for a discussion regarding the need for
 # cache.
-function fmap(f, x; cache = IdDict())
+function fmap(f, x; predicate = isleaf, cache = IdDict())
   haskey(cache, x) && return cache[x]
-  cache[x] = isleaf(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache), x)
+  cache[x] = predicate(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache), x)
```
CarloLucibello, line 43:
```suggestion
  xnew = predicate(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache), x)
  cache[x] = xnew
  return xnew
```
Just a chance to make these rather obscure functions a bit more readable

CarloLucibello (approved):

CarloLucibello:
The name could be shortened to `pred`? `predicate` is fine as well

src/functor.jl
CarloLucibello: actually there is a bug, should pass over the predicate:
```suggestion
  xnew = predicate(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache, predicate = predicate), x)
  cache[x] = xnew
  return xnew
```
Better add tests catching this

README.md
```diff
@@ -71,3 +71,22 @@ Any field not in the list will not be returned by `functor` and passed through a
...
+
+julia> x = ['a', 'b', 'c'];
+
+julia> fmap(cu, x)
+3-element Array{Char,1}:
+ 'a': ASCII/Unicode U+0061 (category Ll: Letter, lowercase)
+ 'b': ASCII/Unicode U+0062 (category Ll: Letter, lowercase)
+ 'c': ASCII/Unicode U+0063 (category Ll: Letter, lowercase)
+
+julia> fmap(cu, x; predicate = CUDA.isbitstype(eltype(x)))
```
ToucheSir, line 87: Should this be a lambda that calls `isbitstype`?

CarloLucibello (approved):

darsnack:
@DhairyaLGandhi any changes or can we merge?

darsnack:
Oh wait I'll rebase tomorrow morning. But leave your comments if you have any, please.

darsnack:
@DhairyaLGandhi I think this is ready for final review

darsnack:
Friendly daily bump @DhairyaLGandhi

PR #11, opened 2021-02-22
Fix extension for compat helper

darsnack:
For some reason the file was named `CompatHelper.jl`


CarloLucibello:
@DhairyaLGandhi can we have write access here? Can we actually have write access for the whole repo, as I have asked countless times?

PR #13, opened 2021-02-23
add fcollect

CarloLucibello:
Follow up to the discussion in https://github.com/FluxML/Flux.jl/pull/1444.

I tried to give maximum flexibility to the interface. 

Notice  that it makes sense to pass a 2-args function `f(v, vs)` instead of a 1-arg `f(v)`, this is because applying a 1-arg `f` is quite simple a-posteriori
```julia
f.(fcollect(m))
```
Possible applications:
- Retrieve leaves from a model:
```julia
filter(isleaf, fcollect(m))
```
- Retrieve all nodes except leaves
```julia
filter(!isleaf, fcollect(m))
# or
fcollect(m; recur = (v, vs) -> !isleaf(v))
```
- Retrieve parents with all leaf children
```julia
a = fcollect(m; f = (v, vs) -> all(isleaf, vs) ? v : nothing)
filter(!isnothing, a)
```

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -29,7 +29,20 @@ macro functor(args...)
   functorm(args...)
 end
 
-isleaf(x) = functor(x)[1] === ()
+"""
+    isleaf(x)
+
+Return true if `x` has no [`children`](@ref) according to [`functor`](@ref).
+"""
+isleaf(x) = children(x) === ()
+
+"""
+    children(x)
+
+Return the children of `x` as defined by [`functor`](@ref).
+Equivalent to `functor(x)[1]`.
+"""
+children(x) = functor(x)[1]
 
 function fmap1(f, x)
   func, re = functor(x)
@@ -65,7 +78,7 @@ function fcollect(x; cache = [],
                      f = (v, vs) -> v)
 
   x in cache && return cache
-  vs = functor(x)[1]
+  vs = children(x)
   recurse(x, vs) || return cache
   push!(cache, f(x, vs))
   foreach(y -> fcollect(y; cache=cache, recurse=recurse, f=f), vs)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -42,3 +42,37 @@ function fmap(f, x; cache = IdDict())
   haskey(cache, x) && return cache[x]
   cache[x] = isleaf(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache), x)
 end
+
+"""
+    fcollect(x; 
+             recurse = (v, vs) -> true, 
+             f = (v, vs) -> v)
+
+Traverse `x` recursively through the children defined by [`functor`](@ref)
+and return an array containing each node encountered.
+
+Doesn't recurse inside branches rooted at nodes `v` with children `vs` 
+for which `recurse(v, vs) == false`.
+In such cases, the root `v` is also excluded from the result.
+Per default, `recurse` yield always true. 
+
+Doesn't recurse inside branches rooted at nodes `v` with children `vs` 
+for which `recurse(v, vs) == false`.
+In such cases, the root `v` is also excluded from the result.
+Per default, `recurse` yield always true. 
+
+A function `f(v, vs)`, taking in input a node and its children, 
+can optionally be passed, so that the returned array will contain
+`f(v, vs)` instead of `v`. 
+"""
+function fcollect(x; cache = [], 
+                     recurse = (v, vs) -> true, 
+                     f = (v, vs) -> v)
+
+  x in cache && return cache
+  vs = functor(x)[1]
+  recurse(x, vs) || return cache
+  push!(cache, f(x, vs))
+  foreach(y -> fcollect(y; cache=cache, recurse=recurse, f=f), vs)
+  return cache
+end
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -57,23 +57,23 @@ function fmap(f, x; cache = IdDict())
 end
 
 """
-    fcollect(x; recurse = v -> true)
+    fcollect(x; exclude = v -> false)
 
 Traverse `x` by recursing each child of `x` as defined by [`functor`](@ref)
 and collecting the results into a flat array.
 
 Doesn't recurse inside branches rooted at nodes `v`
-for which `recurse(v) == false`.
+for which `exclude(v) == true`.
 In such cases, the root `v` is also excluded from the result.
-By default, `recurse` always yields true. 
+By default, `exclude` always yields `false`. 
 
 See also [`children`](@ref).
 """
-function fcollect(x; cache = [], recurse = v -> true)
+function fcollect(x; cache = [], exclude = v -> false)
   x in cache && return cache
-  if recurse(x)
+  if !exclude(x)
     push!(cache, x)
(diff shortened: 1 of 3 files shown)


src/functor.jl
```diff
@@ -42,3 +42,37 @@ function fmap(f, x; cache = IdDict())
...
+can optionally be passed, so that the returned array will contain
+`f(v, vs)` instead of `v`. 
+"""
+function fcollect(x; cache = [], 
+                     recurse = (v, vs) -> true, 
+                     f = (v, vs) -> v)
+
+  x in cache && return cache
+  vs = functor(x)[1]
+  recurse(x, vs) || return cache
```
darsnack, line 82: I think `if recurse(x, vs) ... end` is clearer here

src/functor.jl
```diff
@@ -42,3 +42,32 @@ function fmap(f, x; cache = IdDict())
...
+and return an array containing each node encountered.
+
+Doesn't recurse inside branches rooted at nodes `v` with children `vs` 
+for which `recurse(v, vs) == false`.
+In such cases, the root `v` is also excluded from the result.
+Per default, `recurse` always yields true. 
+
+Optionally, a function `f(v, vs)` taking in input a node and its children 
+can  be passed, so that the returned array will contain
+`f(v, vs)` instead of `v`.
```
darsnack, line 74:
```suggestion
Traverse `x` by recursing each child of `x` as defined by [`functor`](@ref),
applying `f` to each node, and collecting the results into a flat array.

Doesn't recurse inside branches rooted at nodes `v` with children `vs` 
for which `recurse(v, vs) == false`.
In such cases, the root `v` is also excluded from the result.
By default, `recurse` always yields true. 

`f` should accept the inputs `(v, vs)` corresponding to the node
and its children, respectively.
```

darsnack (commented):
I feel like the two-arg version is because we don't have something like `children(x) = functor(x)[1]` in Functors.jl. I would prefer syntax like:
```julia
a = fcollect(m; f = x -> all(isleaf, children(x)) ? x : nothing)
filter(!isnothing, a)
```

CarloLucibello:
> I feel like the two-arg version is because we don't have something like `children(x) = functor(x)[1]` in Functors.jl. I would prefer syntax like:
> 
> ```julia
> a = fcollect(m; f = x -> all(isleaf, children(x)) ? x : nothing)
> filter(!isnothing, a)
> ```

Yeah. I think we can remove the `f` argument altogether and let people handle further processing downstream. 
```julia
filter( x -> all(isleaf, children(x)), fcollect(m))
```
What do you say?

darsnack:
Yeah I think that would match the name `fcollect` better. No reason to include `f` just cause `fmap` has it (which was my original thinking).

darsnack:
We'll want to settle on the same name for `recurse` vs. `predicate`. I think I am okay with `recurse`.

CarloLucibello:
> We'll want to settle on the same name for recurse vs. predicate. I think I am okay with recurse.

I'd go with `recurse`, conveys more intuition, the term `predicate` is very generic. Although the semantic is opposite, `recurse == !predicate`.

test/basics.jl
```diff
@@ -22,17 +27,15 @@ using Functors, Test
...
   model = Baz(1, 2, 3)
   model′ = fmap(x -> 2x, model)
   
   @test (model′.x, model′.y, model′.z) == (1, 4, 3)
 end
+
+
+@testset "Nested" begin
+  model = Bar(Foo(1, [1, 2, 3]))
+  @show fcollect(model)
```
darsnack, line 40:
```suggestion
  @test fcollect(model) == [Bar(Foo(1, [1, 2, 3])), Foo(1, [1, 2, 3]), 1, [1, 2, 3]]
```

darsnack (commented):
Looking good to me. Will probably want @ToucheSir's review since he has more wisdom on Functor-y things than me.

darsnack:
`exclude` is good. Will use that instead of `predicate`.

CarloLucibello:
I changed my mind, i think it is more convenient to specify what to exclude from the recursion instead of what to include, so removed `recurse` in favor of `exclude`.

ToucheSir (approved):
LGTM, this is essentially `Flux.destructure` without trying to shove everything into one vector. One thing we could look at next is creating an iterable version. Then `fcollect` could be expressed simply as `collect(fiter(m, ...))`.

CarloLucibello:
Does anyone have merge rights?

darsnack (approved):

darsnack:
I don't think so

test/basics.jl
```diff
@@ -1,19 +1,24 @@
 using Functors, Test
 
-@testset "Nested" begin
```
DhairyaLGandhi, line 3: Why remove the test set?

src/functor.jl
```diff
@@ -42,3 +55,25 @@ function fmap(f, x; cache = IdDict())
...
+Doesn't recurse inside branches rooted at nodes `v`
+for which `exclude(v) == true`.
+In such cases, the root `v` is also excluded from the result.
+By default, `exclude` always yields `false`. 
+
+See also [`children`](@ref).
+"""
+function fcollect(x; cache = [], exclude = v -> false)
+  x in cache && return cache
+  if !exclude(x)
```
DhairyaLGandhi, line 74: How do you determine what to exclude? Are you suggesting tagging individual arrays somehow?

darsnack: It's a user provided argument. By default it doesn't exclude anything.

DhairyaLGandhi: As I understand it, you can't exclude specific arrays, so maybe we need an MWE on how it's expected to work?

darsnack: > As I understand it, you can't exclude specific arrays

It would be up to the user to figure out some tagging to do this. The main use-case isn't this, but the interface allows for it if that's what the user wants.

test/basics.jl
CarloLucibello: it's not removed. some structs are now defined outside of a specifc testset and used in multiple testsets.

src/functor.jl
CarloLucibello: if you want to exclude a branch rooted at some specific array `a`, you do `exclude = v -> v === a`. If you want to exclude based on some types, you do `exclude = v -> v isa SomeTypeUnion`.

CarloLucibello: > There are some examples in the OP. @CarloLucibello should we put those into the docstring?

yes, I can add some examples to the docstring

CarloLucibello:
we good?

src/functor.jl
```diff
@@ -42,3 +55,58 @@ function fmap(f, x; cache = IdDict())
...
+2-element Vector{Any}:
+ Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
+ Bar([1, 2, 3])
+```
+"""
+function fcollect(x; cache = [], exclude = v -> false)
+  x in cache && return cache
+  if !exclude(x)
+    push!(cache, x)
+    foreach(y -> fcollect(y; cache=cache, exclude=exclude), children(x))
```
DhairyaLGandhi, line 109:
```suggestion
    foreach(y -> fcollect(y; cache = cache, exclude = exclude), children(x))
```

PR #14, opened 2021-07-01
fmapstructure

ToucheSir:
Example usage:
```julia
m = Chain(Dense(10, 5), Dense(5, 2))

functorize(identity, m) |> typeof
# Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Matrix{Float32}, Vector{Float32}, typeof(identity)}}, NamedTuple{(:weight, :bias, :σ), Tuple{Matrix{Float32}, Vector{Float32}, typeof(identity)}}}

# Similar output to what you get from explicit gradients
functorize(x -> x isa AbstractArray ? x : nothing, m) |> typeof
# Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Matrix{Float32}, Vector{Float32}, Nothing}}, NamedTuple{(:weight, :bias, :σ), Tuple{Matrix{Float32}, Vector{Float32}, Nothing}}}
```
Note how the implementation is almost exactly the same as `fmap`, just without the `re`(construction). Ideally this functionality could be folded into `fmap` itself, but I'm not sure what the most elegant way to conditionally control reconstruct is.

cc @DhairyaLGandhi @darsnack

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -49,15 +49,67 @@ function fmap1(f, x)
   re(map(f, func))
 end
 
-# See https://github.com/FluxML/Functors.jl/issues/2 for a discussion regarding the need for
-# cache.
-function fmap(f, x; exclude = isleaf, cache = IdDict())
+"""
+    fmap(f, x; exclude = isleaf, walk = (f′, x) -> ...) 
+    
+A structure and type preserving `map` that works for all [`functor`](@ref)s.
+
+By default, traveres `x` recursively using [`functor`](@ref)
+and transforms every leaf node identified by `exclude` with `f`.
+
+For advanced customization of the traversal behaviour, pass a custom `walk` function.
+This function must itself call the continuation f′ to continue traversal.
+
+# Examples
+```jldoctest
+julia> struct Foo; x; y; end
+
+julia> @functor Foo
+
+julia> struct Bar; x; end
+
+julia> @functor Bar
+
+julia> m = Foo(Bar([1,2,3]), (4, 5));
+
+julia> fmap(x -> 2x, m)
+Foo(Bar([2, 4, 6]), (8, 10))
+
+julia> fmap(string, m)
+Foo(Bar("[1, 2, 3]"), ("4", "5"))
+
+julia> fmap(string, m, exclude = v -> v isa Bar)
+Foo("Bar([1, 2, 3])", (4, 5))
+```
+"""
+function fmap(f, x; exclude = isleaf, walk = fmap1, cache = IdDict())
   haskey(cache, x) && return cache[x]
-  y = exclude(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache, exclude = exclude), x)
+  y = exclude(x) ? f(x) : walk(x -> fmap(f, x, exclude = exclude, walk = walk, cache = cache), x)
   cache[x] = y
 
   return y
 end
+
+"""
+    fmapstructure(f, x; exclude = isleaf)
+    
+Like [`fmap`](@ref), but doesn't preserve the type of custom structs.
+
+Useful for when the output must be plain-old julia data structures.
+
+# Examples
+```jldoctest
+julia> struct Foo; x; y; end
+
+julia> @functor Foo
+
+julia> m = Foo([1,2,3], (4, 5));
+
+julia> fmapstructure(x -> 2x, m)
+(x = [2, 4, 6], y = (8, 10))
+```
+"""
+fmapstructure(f, x; kwargs...) = fmap(f, x; walk=(f, x) -> map(f, children(x)), kwargs...)
 
 """
     fcollect(x; exclude = v -> false)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -50,15 +50,15 @@ function _default_walk(f, x)
 end
 
 """
-    fmap(f, x; exclude = isleaf, walk = (f′, x) -> ...)
+    fmap(f, x; exclude = isleaf, walk = Functors._default_walk)
 
 A structure and type preserving `map` that works for all [`functor`](@ref)s.
 
 By default, traveres `x` recursively using [`functor`](@ref)
 and transforms every leaf node identified by `exclude` with `f`.
 
-For advanced customization of the traversal behaviour, pass a custom `walk` function.
-This function must itself call the continuation f′ to continue traversal.
+For advanced customization of the traversal behaviour, pass a custom `walk` function of the form `(f', xs) -> ...`.
+This function walks (maps) over `xs` calling the continuation `f'` to continue traversal.
 
 # Examples
 ```jldoctest
@@ -93,9 +93,9 @@ end
 """
     fmapstructure(f, x; exclude = isleaf)
 
-Like [`fmap`](@ref), but doesn't preserve the type of custom structs.
+Like [`fmap`](@ref), but doesn't preserve the type of custom structs. Instead, it returns a (potentially nested) `NamedTuple`.
 
-Useful for when the output must be plain-old julia data structures.
+Useful for when the output must not contain custom structs.
 
 # Examples
 ```jldoctest
@@ -109,7 +109,7 @@ julia> fmapstructure(x -> 2x, m)
 (x = [2, 4, 6], y = (8, 10))
 ```
 """
-fmapstructure(f, x; kwargs...) = fmap(f, x; walk=(f, x) -> map(f, children(x)), kwargs...)
+fmapstructure(f, x; kwargs...) = fmap(f, x; walk = (f, x) -> map(f, children(x)), kwargs...)
 
 """
     fcollect(x; exclude = v -> false)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -13,7 +13,7 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
     f in fs ? :(y[$(yᵢ += 1)]) : :(x.$f)
   end
   escfs = [:($f=x.$f) for f in fs]
-  
+
   @eval m begin
     $Functors.functor(::Type{<:$T}, x) = ($(escfs...),), y -> $T($(escargs...))
   end
@@ -44,14 +44,14 @@ Equivalent to `functor(x)[1]`.
 """
 children(x) = functor(x)[1]
 
-function fmap1(f, x)
+function _default_walk(f, x)
   func, re = functor(x)
   re(map(f, func))
 end
 
 """
-    fmap(f, x; exclude = isleaf, walk = (f′, x) -> ...) 
-    
+    fmap(f, x; exclude = isleaf, walk = (f′, x) -> ...)
+
 A structure and type preserving `map` that works for all [`functor`](@ref)s.
 
 By default, traveres `x` recursively using [`functor`](@ref)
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
Nice, this is basically just traversing the tree, right? We could technically replace the construction with an `identity`. maybe an indirection with `fmap1`? I wish we have better names for these functions.

Also, we would miss out on non-functor-able objects in this view which is a shame.

ToucheSir:
Pretty much, I think we'd some kind of kwarg and an additional conditional for fmap, because `re` must be derived at runtime from `x` itself. Then there's the question of whether to pass that kwarg onto `fmap1` and have it handle the conditional, or to make another function that works like `fmap1` but ignores `re`.

Non-functorable objects are handled exactly the same way as they are in `fmap`, `fcollect` etc. Which is to say they're considered leaf nodes and retained in the tree as-is (assuming one uses `exclude=isleaf`).

darsnack:
Is `fmap1` exported? We could just change the name and 0% of users will notice. Could be called `applyfandre`? Or maybe something better than that.

I don't think complicating `fmap` is the way to go. Functors.jl has essentially three functions right now. I would be in favor of just adding a new function like `factorize` instead of more kwargs to `fmap`.

darsnack:
If anything I feel like `f` here serves a similar role to `exclude` in `fcollect`. Would it make more sense to make `fcollect` maintain structure? You could easily flatten the nested tuples after the fact.

ToucheSir:
> I feel like `f` here serves a similar role to `exclude` in `fcollect`

Not really, `exclude` signals when to stop recursing while `f` is a transformation function. Hence `fmap` is really a structure preserving `map`. The reason why `fmap` and `functorize` have so much duplication is because they have different definitions of "structure preserving":

- `fmap` (currently) always maintains the original struct type. i.e. `fmap(identity, m) == m`.
- `functorize` maintains only the structure and values of the original struct.

Come to think of it, we could solve this just by having `fmap1` as a parameter of `fmap`. Then the default function would be `fmap1`, but to replicate `functorize` you'd just pass `(f, x) -> map(f, children(x))`. Now the question becomes what to call that parameter :smile:

darsnack:
Yeah but all the example transforms are just `identity` selectively applied. Well you convinced me anyways, because I think the API that you are proposing is like the "walk" parameter I asked for before. For example, [here](https://github.com/darsnack/FluxPrune.jl/blob/f0896c11ad4fab3f053564f51d9d07951392733a/src/functor.jl#L5-L9) I could use this parameter to only walk the prune-able parameters. So, it's more flexible than just conditional reconstruction.

ToucheSir:
`fmap(length, m)` (for parameter counting) is an example transform that isn't a selectively applied `identity` :)

`walk` isn't a bad name for the parameter. If we're all on the same page about this, I'll switch over the PR to adding that to `fmap`. We could still have a `functorize(f, x; kwargs...) = fmap(f, x; walk=(f, x) -> map(f, children(x)), kwargs...)` as well.

darsnack:
Yeah switching the PR to `walk` sounds good. Maybe we can rename `functorize` to `fmapstructure`?

darsnack:
Maybe add something in the README? Also, I think this can be lifted out of WIP; it's pretty much good to go 😄.

darsnack:
Also should we rename `fmap1` while we're at it?

ToucheSir:
I had to add a better test after finding a bug in the first version (`walk` wasn't being applied recursively). Also added some docstrings based on the `fcollect` docstring.

As for what to name `fmap1`, I'm honestly not sure. `walkfunctor` or `mapfunctor` seem like better names for `fmap` itself honestly. I thought of `destructure_map_restructure`, `fmap_helper`, `default_fmap_helper`, `default_walk`, etc. (all with and without leading underscores), but none of them really stood out.

darsnack:
Well I came up with `walkfunctor`, because I plan to refactor FluxPrune.jl to have `walkpruneable`. Similarly, we could have `walktrainable` one day.

darsnack:
Also, `_fmap` is still better than `fmap1` despite being nearly the same.

darsnack:
After sleeping on it, my vote is for `default_walk`.

ToucheSir:
Thoughts on leading vs no leading underscore? I couldn't find any reference to `fmap1` in the wild, but I imagine we will never want this to leak beyond the internals of Functors.jl.

darsnack:
`_default_walk` and un-exported is probably best to keep it internal.

ToucheSir:
Done. Shall we get this merged so that Flux can start using it?

src/functor.jl
```diff
@@ -44,21 +44,73 @@ Equivalent to `functor(x)[1]`.
...
+"""
+    fmap(f, x; exclude = isleaf, walk = (f′, x) -> ...)
+
+A structure and type preserving `map` that works for all [`functor`](@ref)s.
+
+By default, traveres `x` recursively using [`functor`](@ref)
+and transforms every leaf node identified by `exclude` with `f`.
+
+For advanced customization of the traversal behaviour, pass a custom `walk` function.
+This function must itself call the continuation f′ to continue traversal.
```
darsnack, line 61:
```suggestion
    fmap(f, x; exclude = isleaf, walk = Functors._default_walk)

A structure and type preserving `map` that works for all [`functor`](@ref)s.

By default, traveres `x` recursively using [`functor`](@ref)
and transforms every leaf node identified by `exclude` with `f`.

For advanced customization of the traversal behaviour, pass a custom `walk` function of the form `(f', xs) -> ...`.
This function walks (maps) over `xs` calling the continuation `f'` to continue traversal.
```

darsnack: We should still have an accurate docstring even if the default walk is internal.

src/functor.jl
```diff
@@ -44,21 +44,73 @@ Equivalent to `functor(x)[1]`.
...
+  y = exclude(x) ? f(x) : walk(x -> fmap(f, x, exclude = exclude, walk = walk, cache = cache), x)
   cache[x] = y
 
   return y
 end
 
+"""
+    fmapstructure(f, x; exclude = isleaf)
+
+Like [`fmap`](@ref), but doesn't preserve the type of custom structs.
```
darsnack, line 96:
```suggestion
Like [`fmap`](@ref), but doesn't preserve the type of custom structs. Instead, it returns a (potentially nested) `NamedTuple`.
```

src/functor.jl
```diff
@@ -44,21 +44,73 @@ Equivalent to `functor(x)[1]`.
...
+
+Useful for when the output must be plain-old julia data structures.
```
darsnack, line 98: Maybe we omit this? I'm not sure it's the right wording, since functored structs are still "plain-old Julia data structures."

src/functor.jl
```diff
@@ -44,21 +44,73 @@ Equivalent to `functor(x)[1]`.
...
+
+julia> @functor Foo
+
+julia> m = Foo([1,2,3], (4, 5));
+
+julia> fmapstructure(x -> 2x, m)
+(x = [2, 4, 6], y = (8, 10))
+```
+"""
+fmapstructure(f, x; kwargs...) = fmap(f, x; walk=(f, x) -> map(f, children(x)), kwargs...)
```
darsnack, line 112:
```suggestion
fmapstructure(f, x; kwargs...) = fmap(f, x; walk = (f, x) -> map(f, children(x)), kwargs...)
```

src/functor.jl
```diff
@@ -13,7 +13,7 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
     f in fs ? :(y[$(yᵢ += 1)]) : :(x.$f)
   end
   escfs = [:($f=x.$f) for f in fs]
-  
+
```
darsnack, line 16: A couple of no-op line changes. Do you know what that's from?

darsnack (changes requested):
Left a couple docstring suggestions, but it looks good otherwise.

src/functor.jl
ToucheSir: Not sure, my editor may have trimmed existing trailing whitespace automatically on save?

src/functor.jl
ToucheSir, line 98:
```suggestion
Useful for when the output must not contain custom structs.
```

src/functor.jl
darsnack: Oh yeah, so we'll keep these

src/functor.jl
```diff
@@ -44,21 +44,73 @@ Equivalent to `functor(x)[1]`.
...
 
-# See https://github.com/FluxML/Functors.jl/issues/2 for a discussion regarding the need for
-# cache.
-function fmap(f, x; exclude = isleaf, cache = IdDict())
+"""
+    fmap(f, x; exclude = isleaf, walk = Functors._default_walk)
+
+A structure and type preserving `map` that works for all [`functor`](@ref)s.
+
+By default, traveres `x` recursively using [`functor`](@ref)
```
darsnack, line 57:
```suggestion
By default, traverses `x` recursively using [`functor`](@ref)
```

src/functor.jl
```diff
@@ -44,21 +44,73 @@ Equivalent to `functor(x)[1]`.
...
+
+julia> fmap(x -> 2x, m)
+Foo(Bar([2, 4, 6]), (8, 10))
+
+julia> fmap(string, m)
+Foo(Bar("[1, 2, 3]"), ("4", "5"))
+
+julia> fmap(string, m, exclude = v -> v isa Bar)
+Foo("Bar([1, 2, 3])", (4, 5))
+```
```
darsnack, line 83: Do we want a custom walk example?

darsnack (changes requested):

src/functor.jl
ToucheSir: Woops, thanks!

src/functor.jl
ToucheSir: I couldn't think of a succinct one that wasn't `fmapstructure`, so I just added a doctest for that instead. Are there any examples you can think of?

darsnack: Maybe just something that only walks the children that are arrays of arrays?

ToucheSir: Put in an absolutely uninspired example, but it's an example nonetheless. Happy to change it if something more illustrative comes up.

darsnack (approved):

darsnack:
@CarloLucibello I don't have permissions on this branch. Or do we use `bors`?

DhairyaLGandhi:
Thanks all

DhairyaLGandhi:
The name `fmapstructure` makes it sound like `fmap` itself doesn't preserve structure, which is incorrect. So a better name might be in order.

ToucheSir:
I agree the name isn't ideal, but the intent was to point out that this _only_ preserves the structure (for some handwavy definition of "structure") and not the actual types involved. While `fmap` has signature [`Functor f => (a -> b) -> f a -> f b `](https://hackage.haskell.org/package/base-4.15.0.0/docs/Prelude.html#v:fmap), this would be closer to `Functor f => (a -> b) -> f a -> Either HList HMap` (which is an invalid signature, but my point is to signal the change from a named type into a generic container).

Another way to look at this is as an instance of a poor man's [hylomorphism](https://blog.sumtypeofway.com/posts/recursion-schemes-part-5.html). This take is appealing because it lets you view `fcollect` and `destructure` as catamorphisms (`functor` itself is basically `unfold`), and `_restructure` as an anamorphism (the `re` function returned from `functor` being equivalent to `fold`). The analogy is not perfect because of how different both languages' type systems are, but I think it's a good framework for thinking about the space of operations one could do with a (Functors.jl) functor.

ToucheSir:
Actually, perhaps a more direct analogy would be that `Functors.fmap` with `walk` customization behaves more like [`traverse`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Traversable.html#overview) than `fmap`. In this framing, anything that implements `functor` would be not just a `Functor`, but a [`Traversable`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Traversable.html#g:1). `fmapstructure` would then be closer to calling `traverse` with a function that spits out (named)tuples (themselves `Traversable`s).

PR #17, opened 2021-07-08
add docs to ci and deploy

### LINKED ISSUE #15: Documenter support
Now that we have docstrings for most of the public interface, it would help to have those rendered somewhere. AFAICT we also aren't running any of  the doctests in those docstrings :P

DhairyaLGandhi:
close #15

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,5 @@
 name: CI
-# env:
-#   JULIA_NUM_THREADS: 2
+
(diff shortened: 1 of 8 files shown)


DhairyaLGandhi:
ref https://github.com/FluxML/Flux.jl/pull/1654

We should move be picky but move some of the README into the docs

DhairyaLGandhi:
Cool, this is ready. Now to add some basic literature in here.

PR #19, opened 2021-07-23
FlexibleFunctors

SBuercklin:
This PR is a partial port of [FlexibleFunctors.jl]( https://github.com/Metalenz/FlexibleFunctors.jl), following feedback from @ToucheSir on [this](https://discourse.julialang.org/t/ann-flexiblefunctors-jl-functors-with-per-instance-parameters/64935) Discourse post.

This proposes the `@flexiblefunctor T param_field` syntax for defining a `flexiblefunctor`, or a `functor` with parameters tagged in an internal `Tuple` field, as opposed to a global set of parameters. This enables different instances of a type to have different parameter fields, such as two `Foo` instances where `Foo1.x` is a parameter and `Foo2.y` is a parameter, but `Foo1.y` and `Foo2.x` are not. 

I've reproduced the `@functor` tests (with some modifications to demonstrate flexibility) to demonstrate consistency with the existing framework as well.

This is not a complete 1:1 port from `FlexibleFunctors.jl`. `FlexibleFunctors.jl` actually defines a new function which operates on a type to determine its parameter fields, and for each type this function needed a new method. In practice, we typically store parameters in a field within the struct, so I don't think this feature is a meaningful loss for now.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -32,15 +32,13 @@ end
 function makeflexiblefunctor(m::Module, T, pfield)
   pfield = QuoteNode(pfield)
   @eval m begin
-    $Functors.functor(::Type{<:$T}, x) = begin
+    function $Functors.functor(::Type{<:$T}, x)
+      pfields = getproperty(x, $pfield)
       function re(y)
-        all_args = map(fieldnames($T)) do fn
-          field = fn in getfield(x, $pfield) ? getfield(y, fn) : getfield(x, fn)
-          return field
-        end
+        all_args = map(fn -> getproperty(fn in pfields ? y : x, fn), fieldnames($T))
         return $T(all_args...)
       end
-      func = NamedTuple{pfields}(map(p -> getfield(x, p), pfields))
+      func = NamedTuple{pfields}(map(p -> getproperty(x, p), pfields))
       return func, re
     end
 
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -40,7 +40,7 @@ function makeflexiblefunctor(m::Module, T, pfield)
         end
         return $T(all_args...)
       end
-      func = (; (p => getfield(x, p) for p in getfield(x, $pfield))...)
+      func = NamedTuple{pfields}(map(p -> getfield(x, p), pfields))
       return func, re
     end
 
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -29,6 +29,35 @@ macro functor(args...)
   functorm(args...)
 end
 
+function makeflexiblefunctor(m::Module, T, pfield)
+  pfield = QuoteNode(pfield)
+  @eval m begin
+    $Functors.functor(::Type{<:$T}, x) = begin
+      function re(y)
+        all_args = map(fieldnames($T)) do fn
+          field = fn in getfield(x, $pfield) ? getfield(y, fn) : getfield(x, fn)
+          return field
+        end
+        return $T(all_args...)
+      end
+      func = (; (p => getfield(x, p) for p in getfield(x, $pfield))...)
+      return func, re
+    end
+
+  end
+
+end
+
+function flexiblefunctorm(T, pfield = :params)
+  pfield isa Symbol || error("@flexiblefunctor T param_field")
+  pfield = QuoteNode(pfield)
+  :(makeflexiblefunctor(@__MODULE__, $(esc(T)), $(esc(pfield))))
+end
+
+macro flexiblefunctor(args...)
+  flexiblefunctorm(args...)
+end
+
 """
     isleaf(x)
 
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -2,7 +2,7 @@ module Functors
 
 using MacroTools
 
-export @functor, fmap, fmapstructure, fcollect
+export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect
 
 include("functor.jl")
 
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -73,7 +73,6 @@ end
   m0 = NoChildren(:a, :b)
   m3 = Foo(m2, m0)
   m4 = Bar(m3)
-  println(fcollect(m4))
   @test all(fcollect(m4) .=== [m4, m3, m2, m1, m0])
 end
 
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -76,3 +76,69 @@ end
   println(fcollect(m4))
   @test all(fcollect(m4) .=== [m4, m3, m2, m1, m0])
 end
+
+struct FFoo
+  x
+  y
+  p
+end
+@flexiblefunctor FFoo p
+
+struct FBar
+  x
+  p
+end
+@flexiblefunctor FBar p
+
+struct FBaz
+  x
+  y
+  z
+  p
+end
+@flexiblefunctor FBaz p
+
+@testset "Flexible Nested" begin
+  model = FBar(FFoo(1, [1, 2, 3], (:y, )), (:x,))
+
+  model′ = fmap(float, model)
+
+  @test model.x.y == model′.x.y
+  @test model′.x.y isa Vector{Float64}
+end
+
+@testset "Flexible Walk" begin
+  model = FFoo((0, FBar([1, 2, 3], (:x,))), [4, 5], (:x, :y))
+
+  model′ = fmapstructure(identity, model)
+  @test model′ == (; x=(0, (; x=[1, 2, 3])), y=[4, 5])
+
+  model2 = FFoo((0, FBar([1, 2, 3], (:x,))), [4, 5], (:x,))
+
+  model2′ = fmapstructure(identity, model2)
+  @test model2′ == (; x=(0, (; x=[1, 2, 3])))
+end
+
+@testset "Flexible Property list" begin
+  model = FBaz(1, 2, 3, (:x, :z))
+  model′ = fmap(x -> 2x, model)
+
+  @test (model′.x, model′.y, model′.z) == (2, 2, 6)
+end
+
+@testset "Flexible fcollect" begin
+  m1 = 1
+  m2 = [1, 2, 3]
+  m3 = FFoo(m1, m2, (:y, ))
+  m4 = FBar(m3, (:x,))
+  @test all(fcollect(m4) .=== [m4, m3, m2])
+  @test all(fcollect(m4, exclude = x -> x isa Array) .=== [m4, m3])
+  @test all(fcollect(m4, exclude = x -> x isa FFoo) .=== [m4])
+
+  m0 = NoChildren(:a, :b)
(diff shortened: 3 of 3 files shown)


ToucheSir:
The Discourse announcement has a pretty good rundown of the motivation and why this isn't possible with `@functor` (TL;DR: each instance has a dynamic set of fields).

> Will we be holding on to a lot of memory if we keep fields (many being arrays etc) in a closure separately?

This doesn't do any more holding onto memory than we already do with `@functor Type (fields...)`, which also captures the input if `fields` is a strict subset of `fieldnames(Type)`. Since the mechanism is opt-in, I think it's more than acceptable.

DhairyaLGandhi:
https://github.com/FluxML/Functors.jl/blob/745161772899658c364e8307f1841cb56add743e/src/functor.jl#L13 seems to be the line that can control very similar behavior.

Doesn't the example in the post seem to get it's flexibility from the fact that the `Model` type have a field which marks what is a parameter and what isn't? I guess the idea is to have some kind of runtime information that can be utilised

SBuercklin:
`@functor T (x1, x2, ...)` does not reproduce the same behavior if I want two instances of `T`, one where only `x1` is a parameter and one where only `x2` is a parameter. 

That specific line has `fs` fixed for **all** instances of `T` since it is determined when `@functor` is called. You're right that the `@flexiblefunctor` macro, which is used in a similar fashion to `@functor`, enables *runtime* determination of fields for every instance of `T`. 

Regarding memory, you are correct that this does incur a memory overhead proportional to the `sizeof` all of the non-parameter fields for the given instance of `T` any time you call `Functor.functor(T, x)`. However, these will exist in memory as it stands as the original struct is kept in memory. Arrays should be insignificant since what would be stored is a pointer to the array, rather than a copy of the array itself. Is there a way we could pass a `Ref` to the original object so as to not duplicate these non-parameter fields?

src/functor.jl
```diff
@@ -29,6 +29,35 @@ macro functor(args...)
...
 
+function makeflexiblefunctor(m::Module, T, pfield)
+  pfield = QuoteNode(pfield)
+  @eval m begin
+    $Functors.functor(::Type{<:$T}, x) = begin
+      function re(y)
+        all_args = map(fieldnames($T)) do fn
+          field = fn in getfield(x, $pfield) ? getfield(y, fn) : getfield(x, fn)
+          return field
+        end
```
ToucheSir, line 40:
```suggestion
        all_args = map(fn -> getfield(fn in pfields ? y : x, fn), fieldnames($T))
```

src/functor.jl
```diff
@@ -29,6 +29,35 @@ macro functor(args...)
   functorm(args...)
 end
 
+function makeflexiblefunctor(m::Module, T, pfield)
+  pfield = QuoteNode(pfield)
+  @eval m begin
+    $Functors.functor(::Type{<:$T}, x) = begin
```
ToucheSir, line 35:
```suggestion
    function $Functors.functor(::Type{<:$T}, x)
       pfields = x.$pfield
```
Using `getproperty` instead of `getfield` also leaves the option to override it and add more dynamism to what fields are included (e.g. derived from values of the current fields). My unscientific benchmarking seemed to indicate that hoisting the fields outside of the closure incurs a couple extra allocations, but is ~3x faster.

ToucheSir (commented):
Thanks for this! My comments below:

DhairyaLGandhi:
Maybe we should reflect that in the name? Something that suggests its runtime properties?

src/functor.jl
```diff
@@ -29,6 +29,35 @@ macro functor(args...)
...
+      function re(y)
+        all_args = map(fieldnames($T)) do fn
+          field = fn in getfield(x, $pfield) ? getfield(y, fn) : getfield(x, fn)
+          return field
+        end
+        return $T(all_args...)
+      end
+      func = (; (p => getfield(x, p) for p in getfield(x, $pfield))...)
```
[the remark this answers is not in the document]
DhairyaLGandhi: I was about to suggest the same, but didn't realise it would be that significant.

DhairyaLGandhi:
Structs that want to use runtime information would also be tied to having a field specifically for stating which fields are parameters, which most structs in the Julia world don't really have. 

What I think might be easier is if we have a version of `makefunctor`/ `functor` that accepts an instance as well as fields?

src/functor.jl
```diff
@@ -29,6 +29,35 @@ macro functor(args...)
...
+  @eval m begin
+    $Functors.functor(::Type{<:$T}, x) = begin
+      function re(y)
+        all_args = map(fieldnames($T)) do fn
+          field = fn in getfield(x, $pfield) ? getfield(y, fn) : getfield(x, fn)
+          return field
+        end
+        return $T(all_args...)
+      end
+      func = NamedTuple{pfields}(map(p -> getfield(x, p), pfields))
```
DhairyaLGandhi, line 43:
```suggestion
      func = NamedTuple{pfield}(map(p -> getfield(x, p), pfield))
```

ToucheSir: `pfield` is a symbol. We want the value that symbol indexes to, not the symbol itself.

ToucheSir:
This is why I put in a suggestion to use `getproperty` instead of `getfield` to acquire the list of fields.

There might be a way to integrate this with `makefunctor`, but I can't see a way of doing it without adding essentially 2 independent code paths. In which case having 2 separate functions would be clearer.

DhairyaLGandhi:
Not even with something like `functor(instance, x, fs)`? Rather than defaulting to all fields as in `makefunctor` it can pass the fields explicitly.

src/functor.jl
DhairyaLGandhi: I think ci didn't like it, but of course, we need the value here, so that field needs to be closed over. I'll let @SBuercklin comment on it.

ToucheSir: The reason CI doesn't like it is because it requires the other suggestion https://github.com/FluxML/Functors.jl/pull/19#discussion_r675715573. I didn't want to mark the entire function as one because GH isn't smart enough to handle the diff, but they're technically intertwined.

ToucheSir:
I think I like that idea, but how would a higher-level library like Flux know to call `functor(instance, x, fs)` instead of `functor(instance)`? And where would it look to find `x` and `fs`?

DhairyaLGandhi:
I would imagine that we would need something akin to ```julia
@functor T (x1, x2...)
f, re = Functors.functor(instance, x, fs)
```
I haven't figured out sending the information to Flux just yet, but maybe we need to thread it to `fmap` (which admittedly is getting a bit bloated)

ToucheSir:
Yeah, names aside I think `makeflexiblefunctor` is pretty self contained and any code-sharing would be superficial at best.

SBuercklin:
> I think I like that idea, but how would a higher-level library like Flux know to call `functor(instance, x, fs)` instead of `functor(instance)`? And where would it look to find `x` and `fs`?

We are technically able to store the fields in a tuple in type information and dispatch on that. Part of the problem there is that we need to modify the underlying type parameters and work with custom constructors, and we have a type parameter which is purely metadata and contains no information about the internal fields. 

`@flexiblefunctor` in this case would be an indication that the underlying `Functors.functor` method should look at the type parameters, rather than a field within the type, to extract the tagged fields. 

I worked out something like this at one point, but because you lose your default constructor and had to insert the flagged fields into the type signature, it didn't seem very extensible or easy to work with compared to just inserting a `params` field in my struct.

DhairyaLGandhi:
I think eschewing the field actually makes it more probable to work with existing structs in the Julia world no? Having the types be forwarded to a `functor` method is attractive in that way. Since we can already mark fields at the type level, it should be possible to do so at the instance level as well. We'd be looking into the fields of a struct either way, and still rely on the existence of a default constructor (same as this PR). I'd be interested in what your thoughts are on this. 

This can be future work, but it's good to have a sense of the end game - which to me is optionally having runtime information in the `functor` function.

ToucheSir:
Though handling this at the type level would be nice, I'm not sure adding an additional type parameter to existing structs is any less disruptive than adding an additional field. There's also the question of how this would work in the macro: do we assume that the last type param contains the fields? The first? Ask users to pass in a name and do reflection? 

An alternative could be to pass a function instead of a field name to the macro. Something like `@flexiblefunctor Foo get_params`. The function can then pull from struct fields, type params, an external database and whatever else it wants to generate the list of fields. I was kind of angling at this by suggesting `getproperty` over `getfield`, but having a separate function is likely more efficient than overriding `getproperty`.

SBuercklin:
> I think eschewing the field actually makes it more probable to work with existing structs in the Julia world no? 
No, as all of your types now need to have an extra parametric field in their type definition. In either case, you need to modify your definition of existing types to contain a new field which denotes the tagged fields. In one case, this goes in the type parameters and forces you to only work with custom constructors that handle this extra type parameter. In the other case, you store those parameters in an internal field and the default constructors continue to just work. 

The problem with storing tagging fields in the type information is that it cannot be inferred from the types of the constituent struct fields. This seems like an misuse of the Julia type system to me, and it forces you to manually handle all construction of that type. Working with the tags in the parameters rather than an explicit field is a much messier business.

In all of the `FlexibleFunctors` cases, it requires manual intervention into your type definitions for compatibility. I don't think there is a way to retroactively tag parameters without modifying the types in some way; the information just isn't there. 

> This can be future work, but it's good to have a sense of the end game - which to me is optionally having runtime information in the `functor` function.
I agree that unifying the interface is a great end-goal, and I'd love to see a global syntax that handles everything. What follows is a potential solution, but I would not advocate for it at the moment. It requires a departure from how `Functors.jl` presently works. 

One solution is to operate similar to the `FlexibleFunctors.jl` approach where you define methods for a function which operate on an instance and return the tagged fields. If you do this correctly, you can get methods that are quick and cover all 3 cases now: the original functors, functors with tagged fields stored in a tuple, and functors with tagged fields stored in another type parameter. 

The downside is that this shifts a large burden onto the user to supply methods for the function to determine parameters, and it's a significant departure from the current implementation. It requires both rethinking how `functor(...)` works, potentially a re-write of the syntax for `@functor`, as well as supplying an external method or somehow detecting/automating it internally. At this point, we're doing a really messy re-imagining of `Functors.jl`, and I don't think that's good for users.

SBuercklin:
I believe I've addressed all of the comments in the reviews. Is there anything else I need to take care of for this PR?

ToucheSir (approved):
Think I have to approve to trigger CI, but if this passes I think we're good to merge.

SBuercklin:
Are there any additional changes to make before a merge?

DhairyaLGandhi:
I think we should be good, let me do a final review this week and go from
there?

Best,
Dhairya

DhairyaLGandhi:
Thanks! We should revisit this to implement runtime information without the need of a field, which should get us the benefits regardless, but for now let's have this in

PR #23, opened 2021-08-17
RM MacroTools dep

### LINKED ISSUE #22: Revisit MacroTools Dependency
As far as I can tell, we don't use it anywhere. Removal has the bonus of wider compat and fewer potential errors on nightly.

ToucheSir:
We only used a subset of the functionality from one function in this package. Worth a handful more characters to eliminate an entire dependency here.

Fixes #22

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -20,7 +20,7 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
 end
 
 function functorm(T, fs = nothing)
-  fs === nothing || (fs isa Expr && fs.head === :tuple) || error("@functor T (a, b)")
+  fs === nothing || Meta.isexpr(fs, :tuple) || error("@functor T (a, b)")
   fs = fs === nothing ? [] : [:($(map(QuoteNode, fs.args)...),)]
   :(makefunctor(@__MODULE__, $(esc(T)), $(fs...)))
 end
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -20,8 +20,8 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
 end
 
 function functorm(T, fs = nothing)
-  fs == nothing || isexpr(fs, :tuple) || error("@functor T (a, b)")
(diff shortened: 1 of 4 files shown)


src/functor.jl
```diff
@@ -20,8 +20,8 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
 end
 
 function functorm(T, fs = nothing)
-  fs == nothing || isexpr(fs, :tuple) || error("@functor T (a, b)")
-  fs = fs == nothing ? [] : [:($(map(QuoteNode, fs.args)...),)]
+  fs === nothing || (fs isa Expr && fs.head === :tuple) || error("@functor T (a, b)")
```
oxinabox, line 23:
```suggestion
  fs === nothing || (Meta.isexpr(fs, :tuple) || error("@functor T (a, b)")
```
I think works

oxinabox (approved):

src/functor.jl
ToucheSir: It does indeed, thanks!

src/functor.jl
```diff
@@ -20,8 +20,8 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
 end
 
 function functorm(T, fs = nothing)
-  fs == nothing || isexpr(fs, :tuple) || error("@functor T (a, b)")
-  fs = fs == nothing ? [] : [:($(map(QuoteNode, fs.args)...),)]
+  fs === nothing || Meta.isexpr(fs, :tuple) || error("@functor T (a, b)")
+  fs = fs === nothing ? [] : [:($(map(QuoteNode, fs.args)...),)]
```
darsnack, line 24:
```suggestion
  isnothing(fs) || Meta.isexpr(fs, :tuple) || error("@functor T (a, b)")
  fs = isnothing(fs) ? [] : [:($(map(QuoteNode, fs.args)...),)]
```

darsnack: Ignore, thought the `===` was causing some parsing issue.

DhairyaLGandhi:
Kind of a light dep but good to reduce some deps 👍

PR #24, opened 2021-09-08
Support Base.ComposedFunction

### LINKED ISSUE #21: Support Base.ComposedFunction
Given

```julia
using Functors

struct Foo <: Function
    a::Float64
end

(f::Foo)(x) = f.a * x

@functor Foo
```

it would IMHO be very useful to support `ComposedFunction`, to enable

```julia
Functors.functor(Foo(4.2) ∘ Foo(5.3)) == ((outer = (a = 4.2,), inner = (a = 5.3,)), re)
```

If acceptable, I'd volunteer to implement it.

*Update: Fixed example above*

oschulz:
Closes #21.

### DIFF
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -23,6 +23,16 @@ struct NoChildren
   y
 end
 
+@static if VERSION >= v"1.6"
+  @testset "ComposedFunction" begin
+    f1 = Foo(1.1, 2.2)
+    f2 = Bar(3.3)
+    @test Functors.functor(f1 ∘ f2)[1] == (outer = f1, inner = f2)
+    @test Functors.functor(f1 ∘ f2)[2]((outer = f1, inner = f2)) == f1 ∘ f2
+    @test fmap(x -> x + 10, f1 ∘ f2) == Foo(11.1, 12.2) ∘ Bar(13.3)
+  end
+end
+
 @testset "Nested" begin
   model = Bar(Foo(1, [1, 2, 3]))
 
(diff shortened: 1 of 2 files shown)


oschulz:
I also tested this in a small use case with Optimisers.jl (using Zygote gradients), works nicely.

oschulz:
@DhairyaLGandhi / @darsnack, could you approve the workflow?

test/basics.jl
```diff
@@ -23,6 +23,14 @@ struct NoChildren
   y
 end
 
+@testset "ComposedFunction" begin
+  f1 = Foo(1.1, 2.2)
+  f2 = Bar(3.3)
+  @test Functors.functor(f1 ∘ f2)[1] == (outer = f1, inner = f2)
+  @test Functors.functor(f1 ∘ f2)[2]((outer = f1, inner = f2)) == f1 ∘ f2
+  fmap(x -> x + 10, f1 ∘ f2) == Foo(11.1, 12.2) ∘ Bar(13.3)
```
DhairyaLGandhi, line 31:
```suggestion
  @test fmap(x -> x + 10, f1 ∘ f2) == Foo(11.1, 12.2) ∘ Bar(13.3)
```

oschulz: Whoops, thanks!

DhairyaLGandhi:
can we add a version check for 1.6?

oschulz:
> can we add a version check for 1.6?

Yep, just did. Forgot that `ComposedFunction` wasn't there in v1.5 yet.

DhairyaLGandhi:
Lgtm, thanks! We would prolly want a way to handle tuples in the restructure function but I suspect it's not come up yet. I'll merge it for now.

oschulz:
Thanks! Could you do a v0.2.5 release with this? I'd like to use it in https://github.com/bat/EuclidianNormalizingFlows.jl .

PR #25, opened 2021-10-05
Fix `fcollect` to respect object identity instead of `==`; document order

### LINKED ISSUE #16: `fcollect` won't keep repeated arrays that are `==` but distinct
See <https://github.com/beacon-biosignals/LegolasFlux.jl/pull/4#issuecomment-875546970> and the alternate implementation
```julia
function fcollect2(x; output = [], cache = Base.IdSet(), exclude = v -> false)
    x in cache && return output
    if !exclude(x)
      push!(cache, x)
      push!(output, x)
      foreach(y -> fcollect2(y; cache = cache, output=output, exclude = exclude), Functors.children(x))
    end
    return output
end
```

ericphanson:
Fixes #16. I checked the test I added failed on master:
```julia
julia> m1 = [1, 2, 3]
3-element Vector{Int64}:
 1
 2
 3

julia> m2 = [1, 2, 3]
3-element Vector{Int64}:
 1
 2
 3

julia> m3 = Foo(m1, m2)
Foo([1, 2, 3], [1, 2, 3])

julia> fcollect(m3)
2-element Vector{Any}:
 Foo([1, 2, 3], [1, 2, 3])
 [1, 2, 3]

julia> @test all(fcollect(m3) .=== [m3, m1, m2])
Error During Test at REPL[8]:1
  Test threw exception
  Expression: all(fcollect(m3) .=== [m3, m1, m2])
  DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 2 and 3")
```

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -192,11 +192,12 @@ julia> fcollect(m, exclude = v -> Functors.isleaf(v))
  Bar([1, 2, 3])
 ```
 """
-function fcollect(x; cache = [], exclude = v -> false)
-  x in cache && return cache
-  if !exclude(x)
-    push!(cache, x)
-    foreach(y -> fcollect(y; cache = cache, exclude = exclude), children(x))
-  end
-  return cache
+function fcollect(x; output = [], cache = Base.IdSet(), exclude = v -> false)
+    x in cache && return output
+    if !exclude(x)
+      push!(cache, x)
+      push!(output, x)
+      foreach(y -> fcollect(y; cache=cache, output=output, exclude=exclude), children(x))
+    end
+    return output
 end
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -149,7 +149,8 @@ fmapstructure(f, x; kwargs...) = fmap(f, x; walk = (f, x) -> map(f, children(x))
     fcollect(x; exclude = v -> false)
 
 Traverse `x` by recursing each child of `x` as defined by [`functor`](@ref)
-and collecting the results into a flat array.
+and collecting the results into a flat array, ordered by a breadth-first
+traversal of `x`, respecting the iteration order of `children` calls.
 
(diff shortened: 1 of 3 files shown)


darsnack (approved):
Looks correct to me!

src/functor.jl
```diff
@@ -192,11 +192,12 @@ julia> fcollect(m, exclude = v -> Functors.isleaf(v))
...
 """
-function fcollect(x; cache = [], exclude = v -> false)
-  x in cache && return cache
-  if !exclude(x)
-    push!(cache, x)
-    foreach(y -> fcollect(y; cache = cache, exclude = exclude), children(x))
-  end
-  return cache
+function fcollect(x; output = [], cache = Base.IdSet(), exclude = v -> false)
+    x in cache && return output
```
ericphanson, line 196: I realized it wasn't totally obvious why the code was done this way and also that LegolasFlux will depend on the order (for the same model) not changing in non-breaking releases of Functors.jl, so maybe this comment will help communicate that:
```suggestion
    # note: we don't have an `OrderedIdSet`, so we use an `IdSet` for the cache
    # (to ensure we get exactly 1 copy of each distinct array), and a usual `Vector`
    # for the results, to preserve traversal order (important downstream!).
    x in cache && return output
```
Note the tests already test the order, so hopefully that will be another sign to not break it 😄.

DhairyaLGandhi:
If we intend to preserve the order, its worth a mention in the docs.

ericphanson:
Ok, I wrote up the order defined by the implementation in the docstring. We could also not say exactly what the order is but just say it won't change for any inputs in a non-breaking release-- would that be preferred?

darsnack (approved):

darsnack:
I think the current explanation is more precise and understandable.

PR #26, opened 2021-10-12
functor RefValue

CarloLucibello:
related to https://github.com/FluxML/Flux.jl/issues/1727

### DIFF
--- a/test/base.jl
+++ b/test/base.jl
@@ -1,7 +1,7 @@
 @testset "Base" begin
     @testset "RefValue" begin
         x = Ref(1)
-        p, re = functor(x)
+        p, re = Functors.functor(x)
         @test p == (x = 1,)
         @test re(p) isa Base.RefValue{Int}
     end
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -1,5 +1,3 @@
-using Functors, Test
-
 struct Foo
   x
   y


DhairyaLGandhi:
No, I don't think so. Our options for defining leaves are either them being arrays or them not having been `functor`ed. Writing filtered walks are not always easy to implement anyway. So having a convenient way which already works seems sensible. We have a fair few options to handle #1727 as mentioned in the issue comments.

darsnack:
If the argument is that you can make something a leaf by wrapping it in `Ref`, then I think that re-using `Ref` is API abuse. `Ref` can show up in arbitrary Julia structures and implicitly assuming it has some fixed meaning is not flexible design. If we want a wrapper type that behaves like `Ref` but is understood as leaf by Functors.jl, then create a `Leaf` type. It will have one unambiguous meaning.

Also, if a certain type that is not an array should always be considered a leaf, then the author of that type can always extend `Functors.functor` to make it leaf anywhere it is used. Walks are not meant to be a mechanism to declaring leaves. It is mechanism for dynamically choosing how to walk over _all_ leaves (i.e. making it possible for external APIs like `trainable` to compose well with `fmap`).

DhairyaLGandhi:
Seems like having something closer to https://github.com/FluxML/Flux.jl/issues/1727#issuecomment-931749324 would be the correct implementation. We can discuss the semantic function Ref plays for us, but it is not only mutables that can produce a Ref within Zygote.

ToucheSir:
There's still the problem of distinguishing a `Ref` that was already in the model from one created by Zygote. Unfortunately, it's hard to say what the correct semantics should be since there doesn't seem to be AD support ATM:

```julia
julia> struct HasRef{T}
         inner::RefValue{T}
       end

julia> gradient(hr -> hr.inner[] * 5, HasRef(Ref(1)))
(nothing,)

julia> gradient(x -> x[] * 5, Ref(1))
ERROR: MethodError: no method matching (::ChainRulesCore.ProjectTo{Float64, NamedTuple{(), Tuple{}}})(::NamedTuple{(:x,), Tuple{Float64}})
Closest candidates are:
  (::ChainRulesCore.ProjectTo{T, D} where D<:NamedTuple)(::ChainRulesCore.AbstractZero) where T at /home/brianc/.julia/packages/ChainRulesCore/8vlYQ/src/projection.jl:120
  (::ChainRulesCore.ProjectTo{var"#s15", D} where {var"#s15"<:Real, D<:NamedTuple})(::Complex) at /home/brianc/.julia/packages/ChainRulesCore/8vlYQ/src/projection.jl:179
  (::ChainRulesCore.ProjectTo{var"#s15", D} where {var"#s15"<:Number, D<:NamedTuple})(::ChainRulesCore.Tangent{var"#s14", T} where {var"#s14"<:Number, T}) at /home/brianc/.julia/packages/ChainRulesCore/8vlYQ/src/projection.jl:185
  ...
Stacktrace:
 [1] (::ChainRulesCore.ProjectTo{Ref, NamedTuple{(:type, :x), Tuple{DataType, ChainRulesCore.ProjectTo{Float64, NamedTuple{(), Tuple{}}}}}})(dx::Base.RefValue{Any})
   @ ChainRulesCore ~/.julia/packages/ChainRulesCore/8vlYQ/src/projection.jl:275
 [2] _project
   @ ~/.julia/packages/Zygote/EPhp6/src/compiler/chainrules.jl:140 [inlined]
 [3] map(f::typeof(Zygote._project), t::Tuple{Base.RefValue{Int64}}, s::Tuple{Base.RefValue{Any}})
   @ Base ./tuple.jl:232
 [4] gradient(f::Function, args::Base.RefValue{Int64})
   @ Zygote ~/.julia/packages/Zygote/EPhp6/src/compiler/interface.jl:77
 [5] top-level scope
   @ REPL[8]:1
```

ToucheSir:
You're right, I'd forgotten that the whole broader conversation about Refs happened later in that thread. Still, I think the best compromise would be to only functor the Ref types we expect to receive from Zygote (side note: a list of these would be very much appreciated) and no more. Forwarding the functor call like I described in my last comment would also allow us to remove some similar special-cased code in Optimisers.jl.

CarloLucibello:
Zygote returns `RefValue{Any}` on the cases I tested. It seems a bit odd to functorize only `RefValue{Any}`, and I'm not sure it covers all Zygote can return. Also, as expressed in https://github.com/FluxML/Flux.jl/issues/1741, I think we should start being more bold with what we functorize, and this is a very tiny step in that direction.

ToucheSir:
I don't necessarily disagree, but doing this just made work for Optimisers.jl harder. It seems like mutable struct Refs may have been addressed by https://github.com/FluxML/Zygote.jl/pull/1102, however, so we'll have to give it all a test.

PR #27, opened 2021-10-14
Functorizing `Functors`

ToucheSir:
Since there's been a lot of talk about the current interface and limitations thereof, I thought I'd clean up and post a experimental branch that seeks to straighten out the library and make it more flexible. For those of you who saw me talking about recursion schemes and whatnot, this is what that was for ;)

The tests contain examples from this [blog series](https://blog.sumtypeofway.com/posts/recursion-schemes-part-6.html) as well as real-world problems that have come up before in Flux (e.g. https://github.com/FluxML/Flux.jl/issues/1284). I plan on adding some Optimisers.jl-related ones and more as this develops. Obviously everything is very, very breaking, so if the experiment works out we'll have to think about potential migration plans. But until then, feedback and use cases where Functors.jl currently fails (e.g. when a custom walk function is required) are welcome!

### DIFF
--- a/old_src/Functors.jl
+++ b/old_src/Functors.jl
@@ -1,0 +1,7 @@
+module Functors
+
+export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect
+
+include("functor.jl")
+
+end # module
--- a/old_src/functor.jl
+++ b/old_src/functor.jl
@@ -1,0 +1,202 @@
+functor(T, x) = (), _ -> x
+functor(x) = functor(typeof(x), x)
+
+functor(::Type{<:Tuple}, x) = x, y -> y
+functor(::Type{<:NamedTuple}, x) = x, y -> y
+
+functor(::Type{<:AbstractArray}, x) = x, y -> y
+functor(::Type{<:AbstractArray{<:Number}}, x) = (), _ -> x
+
+@static if VERSION >= v"1.6"
+  functor(::Type{<:Base.ComposedFunction}, x) = (outer = x.outer, inner = x.inner), y -> Base.ComposedFunction(y.outer, y.inner)
+end
+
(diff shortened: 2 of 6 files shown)


CarloLucibello:
Can you summarize what this PR does?

ToucheSir:
There were two main motivations for this PR:

1. We're constantly having to override `_walk` in `fmap` for custom functions, if not ignore it altogether. This should be low-level internal functionality, so such frequent usage suggests the need for a better interface/abstraction.
2. Functors.jl currently allocates a lot of closures. What's more, you have to hold onto these closures to reconstruct many types because they capture non-functored fields as well. What ChainRulesCore showed with `Tangent` and `ProjectTo` is that reifying the transformed value into its own type can help quite a bit here. Yes it's nice to have a plain NamedTuple, but that has no additional semantic meaning which would allow you to tie it back to the original type it was created from.

One of the underlying issues with the current implementation is that the "functors" aren't really `Functor`s in an FP sense. `Functors.functor` works fine for flattening, but my understanding is the goal was always to support generalized tree (well, DAG) traversals. Otherwise, the value proposition of Functors.jl over https://github.com/rafaqz/Flatten.jl becomes somewhat dubious. Thus, this PR aims to address any instance where we've had to work around Functors.jl functions in order to make something work. That includes every open Functors-related PR in this repo, Flux or Optimisers.jl. I won't belabour the point enumerating every instance, so if you're curious about a specific one then we can discuss from there.

ToucheSir:
In the spirit of not keeping too many Zombie PRs around, this was an interesting experiment but is unlikely to become reality any time soon. In particular, the programming interface is a little too "un-julian". Some of the interesting bits are also already in Functors or Optimisers as separate commits/PRs.

PR #30, opened 2022-01-22
Expand docstrings

mcabbott:
This adds more examples to show what this package actually does. Mostly examples, but I fixed the words where I thought them ambiguous, added something about `===` etc.

Should have no functional changes.

Documenter will run doctests when you do `] test`, not only when building docs.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -1,3 +1,13 @@
+"""
+    Functors.functor(x) = functor(typeof(x), x)
+
+Returns a tuple containing, first, a `NamedTuple` of the children of `x`
+(typically its fields), and second, a reconstruction funciton.
+This controls the behaviour of [`fmap`](@ref).
+
+Methods should be added to `functor(::Type{T}, x)` for custom types,
+usually using the macro [@functor](@ref).
+"""
 functor(T, x) = (), _ -> x
 functor(x) = functor(typeof(x), x)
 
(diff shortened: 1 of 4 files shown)


src/functor.jl
```diff
@@ -100,18 +172,55 @@ end
...
 
 """
-    fmap(f, x; exclude = isleaf, walk = Functors._default_walk)
+    fmap(f, x; exclude = Functors.isleaf, walk = Functors._default_walk)
 
-A structure and type preserving `map` that works for all [`functor`](@ref)s.
+A structure and type preserving `map`.
 
-By default, traverses `x` recursively using [`functor`](@ref)
-and transforms every leaf node identified by `exclude` with `f`.
```
mcabbott, line 108: It's important to know that `exclude` is used before `functor`, it stops the recursion, rather than just controlling whether `f` is applied. I'm not sure that's what we want, but it's what we have.

ToucheSir: Yes, this is deliberate because it's the easiest way to control the depth of recursion. Looking back on https://github.com/FluxML/Functors.jl/pull/13, I think it might've been better just to call this `isleaf`: that's literally what it's checking and `exclude` could imply that `f` is _not_ being applied to anything the function predicate says is true.

ToucheSir (approved):

src/functor.jl
mcabbott: Ok. Maybe all names for such things are confusing... 

What it sounds like it might do is let you tell `fmap(sqrt, (2, Foo(3, sin)))` which leaves to apply `f` to, and which to leave alone. I guess you can test `isleaf` first. Maybe all functionality can be written in terms of any other...

ToucheSir: It's a little confusing because `f` is also used as a name for not the original user callback, but in essence Functors will try to recurse into anything "left alone" by `isleaf`/`exclude`. It's just that the fallback returns an empty tuple of children, so recursion naturally stops if the value in question can't be unpacked.

Issue #28, opened 2022-01-14
Add more Base types


mcabbott:
This should just work, right? https://discourse.julialang.org/t/data-science-lessons-making-10-neural-networks-run-on-gpu/74592

darsnack:
This is a bit of a drive-by comment, some of this might already be addressed. See some earlier discussion in #21. I think in general this would make sense to do, but we need to think carefully about what should be a lead (i.e. not functor'd) and the ergonomics of users overriding something Functors.jl says is not a leaf to be a leaf.

mcabbott:
Thanks, I didn't see that, have not followed this package closely.

I'm not actually too sure why it requires types to be annotated, rather than the reverse, of recursing into anything unless told not to. That would seem more convenient at least for what Flux wants. 

I guess you have the potential to run into things without default constructors... but at least that's a noisy failure not a silent don't-do-what-I-asked. I thought Setfield worked around this but it seems I was mistaken:
```
julia> using Setfield

julia> struct Foo{T,S} x::T; y::S end;

julia> foo = Foo(1,2); @set foo.x = "not an integer"
Foo{String, Int64}("not an integer", 2)

julia> struct Bar{T,S} x::T; y::S;
       Bar(x::T) where T = new{T,T}(x)
       end;

julia> bar = Bar(1); @set bar.x = "not an integer"
ERROR: MethodError: no method matching Bar(::String, ::Int64)
```

ToucheSir:
Setfield.jl relies on ConstructionBase.jl, which has the benefit of being a standardized interface but is not fancy (read: crazy) enough to make `Bar` work automatically. We could try emulating what BSON.jl does to resuscitate types during deserialization, but the level of black magic required there has caused more than one headache in the past.

mcabbott:
Ok. Without going so far, the desirable Base types to handle seem to be
* Iterators, like the above issue. These all have default constructors. 
* Transpose, Adjoint, and maybe ReshapedArray, PermutedDimsArray. For the reason that shared weights will often be transposed.

Shared weights with Transpose seem somewhat tricky, since the gradient of a Transpose will often be a Matrix, so you need to know the inverse transformation. But not impossible, seems a smaller change than (say) running the `cache` based on `pointer`?

Is it a breaking change to add types to recurse into?

darsnack:
> Transpose, Adjoint, and maybe ReshapedArray, PermutedDimsArray. For the reason that shared weights will often be transposed.

[...]

PR #31, opened 2022-01-23
`fvec`

mcabbott:
This is an attempt to write a function which does what `Flux.destructure` does, but more carefully. It allows for missing branches of the structural gradient, and for shared weights having separate gradients, which ought to accumulate in making a vector. When returning to a structural gradient, only the first node gets the gradient, to avoid mistakenly doubling it.

It also declares some Base types like Transpose to be functors, since a decoder with transposed weights seems like the canonical example. And sets up a way to allow their gradients not to be Transpose.

And, thirdly, it changes `fmap` to not use the cache on `isbits` objects. This is to avoid falsely tying two parameters initially say `SA[0,0,0]`. We thought elsewhere that the way to indicate that you do what such things tied is to wrap them in some `TiedArray` type like https://gist.github.com/mcabbott/35f660ff9fb8e7b0b23a2abb94618cf4, but not in this PR.

---

Most of this can simply run with a different walk to hit only trainable arrays. Which I think ought to be done in Optimisers.jl, which owns `trainable`. The exception is that the gradient of `fcopy` doesn't seem to fit into just using a different walk, maybe there's a way? It takes a different `children` function instead.

Not sure what should be exported here, or what anything should be called. I think the name `destructure` should belong to Optimisers.jl too. Perhaps this package should export a function which similarly returns a vector and a reconstructor. That's just one line after the pieces defined here. (But easier to write & test the pieces separately.)

---

Needs more tests, still.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -1,1 +1,24 @@
+
 @functor Base.RefValue
+
+functor(::Type{<:Base.ComposedFunction}, x) = (outer = x.outer, inner = x.inner), y -> Base.ComposedFunction(y.outer, y.inner)
+
+functor(::Type{<:PermutedDimsArray{T,N,perm}}, x) where {T,N,perm} = (parent = parent(x),), y -> PermutedDimsArray(only(y), perm)
+
+using Base.Iterators
+
+Functors.@functor Iterators.Drop (xs,)
+Functors.@functor Iterators.Enumerate
+# Functors.@functor Iterators.Flatten
+# Functors.@functor Iterators.ProductIterator
+# Functors.@functor Iterators.PartitionIterator
+Functors.@functor Iterators.Repeated
+Functors.@functor Iterators.Reverse
+Functors.@functor Iterators.Take (xs,)
+Functors.@functor Iterators.Zip
+
+using LinearAlgebra
+
+@functor Adjoint
+@functor Transpose
+@functor Diagonal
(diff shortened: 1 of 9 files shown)


src/base.jl
```diff
@@ -1 +1,38 @@
...
+using LinearAlgebra
+# The reason for these is to let W and W' be seen as tied weights in Flux models.
+# But the problem is that the gradient of W' may be a Matrix, so the trees don't match,
+# so we will call `lazywrap` instead of walking:
+
+functor(::Type{<:Adjoint}, x) = (parent = parent(x),), y -> adjoint(only(y))
+lazywrap(x::Adjoint) = parent(x), adjoint
+
+functor(::Type{<:Transpose}, x) = (parent = parent(x),), y -> transpose(only(y))
+lazywrap(x::Transpose) = parent(x), transpose
```
mcabbott, line 29: This is #28 and should perhaps be pulled out. 

From https://github.com/FluxML/Optimisers.jl/pull/42 I think the right pattern is actually more like this -- calling `parent(x)` is not safe, and we may want to transpose things which aren't matrix-like.
```
Functors.functor(::Type{<:Transpose}, x) = (parent = _transpose(x),), y -> _transpose(only(y))
_transpose(x) = transpose(x)
_transpose(x::NamedTuple{(:parent,)}) = x.parent  # "structural" gradient
# _transpose(bc::Broadcast.Broadcasted{S}) where S = Broadcast.Broadcasted{S}(map(_transpose, bc.args)) do args...
#     transpose(bc.f(map(transpose, args)...))
# end
_transpose(bc::Broadcast.Broadcasted) = transpose(Broadcast.materialize(bc))
```

PR #32, opened 2022-01-30
Extract common functionality into fold

ToucheSir:
Traversal functions in Functors.jl currently handle traversal via manual recursion. This isn't the end of the world, but it results in a good amount of code duplication as well as parameter explosion from having to plumb auxiliary arguments through the call stack. With #31 looking to add a number of additional functions and Optimisers.jl not getting a lot of mileage out of `fmap`, it's a good time to consider whether we can cut down on the boilerplate.

This PR introduces helper traversal functionality in the form of `[fF]old`. In the language of [recursion schemes](https://blog.sumtypeofway.com/posts/introduction-to-recursion-schemes.html), `fold` is a "catamorphism" which performs a generalized structural reduction over a tree (or in our case, DAG).

This is the first step in an effort to implement the vision of #27 while maintaining as much backwards compatibility as possible. A non-exhaustive list of objectives for future PRs include:
1. Reducing the reliance on custom walks for downstream code
2. Removing the requirement to carry around the `re`(construct) closure. This can lead to some unfortunate gymnastics, so getting rid of it would be great.
3. Accomodating multiple inputs, e.g. `fmap(f, x, xs...)`. This would help Optimisers.jl and any other downstream library that have rolled their own `fmap` variants with `functor`.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -26,7 +26,7 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
   escargs = map(fieldnames(T)) do f
     f in fs ? :(y[$(yᵢ += 1)]) : :(x.$f)
   end
-  escfs = [:($f=x.$f) for f in fs]
+  escfs = [:($f = x.$f) for f in fs]
 
   @eval m begin
     $Functors.functor(::Type{<:$T}, x) = ($(escfs...),), y -> $T($(escargs...))
@@ -169,7 +169,52 @@ function _default_walk(f, x)
   func, re = functor(x)
   re(map(f, func))
 end
-_default_walk(f, ::Nothing, ::Nothing) = nothing
+_default_walk(_, ::Nothing, ::Nothing) = nothing
+
+# Side effects only, saves a restructure
+function _foreach_walk(f, x)
+  foreach(f, children(x))
+  return x
+end
+
+### WARNING: the following is unstable internal functionality. Use at your own risk!
+# Wrapper over an IdDict which only saves values with a stable object identity
+struct Cache{K,V}
+  inner::IdDict{K,V}
+end
+Cache() = Cache(IdDict())
+
+iscachesafe(x) = !isbits(x) && ismutable(x)
+# Functionally immutable and observe value semantics, but still `ismutable` and not `isbits`
+iscachesafe(::Union{String,Symbol}) = false
+# For varargs
+iscachesafe(xs::Tuple) = all(iscachesafe, xs)
+Base.get!(f, c::Cache, x) = iscachesafe(x) ? get!(f, c.inner, x) : f(x)
+
+# Passthrough used to disable caching (e.g. when passing `cache=false`)
+struct NoCache end
+Base.get!(f, ::NoCache, x) = f(x)
+
+# Encapsulates the self-recursive part of a recursive tree reduction (fold).
+# This allows calling functions to remove any self-calls or nested callback closures.
+struct Fold{F,L,C,W}
+  fn::F
+  isleaf::L
+  cache::C
+  walk::W
+end
+(fld::Fold)(x) = get!(() -> fld.fn(fld.isleaf(x) ? x : fld.walk(fld, x)), fld.cache, x)
+
+# Convenience function for working with `Fold`
+function fold(f, x; isleaf = isleaf, cache = false, walk = _default_walk)
+  if cache === true
+    cache = Cache()
+  elseif cache === false
+    cache = NoCache()
+  end
+  return Fold(f, isleaf, cache, walk)(x)
+end
+### end of unstable internal functionality
 
 """
     fmap(f, x; exclude = Functors.isleaf, walk = Functors._default_walk)
@@ -253,11 +298,10 @@ Foo(Bar([1, 2, 3]), (40, 50, Bar(Foo(6, 7))))
 ```
 """
 function fmap(f, x; exclude = isleaf, walk = _default_walk, cache = IdDict())
-  haskey(cache, x) && return cache[x]
-  y = exclude(x) ? f(x) : walk(x -> fmap(f, x, exclude = exclude, walk = walk, cache = cache), x)
-  cache[x] = y
-
-  return y
+  return fold(x; cache, walk, isleaf = exclude) do node
+    !exclude(node) && return node
+    return f(node)
+  end
 end
 
 """
@@ -296,8 +340,8 @@ fmapstructure(f, x; kwargs...) = fmap(f, x; walk = (f, x) -> map(f, children(x))
     fcollect(x; exclude = v -> false)
 
 Traverse `x` by recursing each child of `x` as defined by [`functor`](@ref)
-and collecting the results into a flat array, ordered by a breadth-first
-traversal of `x`, respecting the iteration order of `children` calls.
+and collecting the results into a flat array, ordered by a depth-first,
+post-order traversal of `x` that respects the iteration order of `children` calls.
 
 Doesn't recurse inside branches rooted at nodes `v`
 for which `exclude(v) == true`.
@@ -324,33 +368,27 @@ Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
 
 julia> fcollect(m)
 4-element Vector{Any}:
- Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
- Bar([1, 2, 3])
  [1, 2, 3]
+ Bar([1, 2, 3])
  NoChildren(:a, :b)
+ Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
 
 julia> fcollect(m, exclude = v -> v isa Bar)
 2-element Vector{Any}:
- Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
  NoChildren(:a, :b)
+ Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
 
 julia> fcollect(m, exclude = v -> Functors.isleaf(v))
 2-element Vector{Any}:
- Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
  Bar([1, 2, 3])
+ Foo(Bar([1, 2, 3]), NoChildren(:a, :b))
 ```
 """
-function fcollect(x; output = [], cache = Base.IdSet(), exclude = v -> false)
-    # note: we don't have an `OrderedIdSet`, so we use an `IdSet` for the cache
-    # (to ensure we get exactly 1 copy of each distinct array), and a usual `Vector`
-    # for the results, to preserve traversal order (important downstream!).
-    x in cache && return output
-    if !exclude(x)
-      push!(cache, x)
-      push!(output, x)
-      foreach(y -> fcollect(y; cache=cache, output=output, exclude=exclude), children(x))
-    end
-    return output
+function fcollect(x; output = [], cache = Base.IdDict(), exclude = v -> false)
+  fold(x; cache, isleaf = exclude, walk = _foreach_walk) do node
+    exclude(node) || push!(output, node);  # always return nothing
+  end
+  return output
 end
 
 # Allow gradients and other constructs that match the structure of the functor
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -72,8 +72,8 @@ end
   m2 = 1
   m3 = Foo(m1, m2)
   m4 = Bar(m3)
-  @test all(fcollect(m4) .=== [m4, m3, m1, m2])
-  @test all(fcollect(m4, exclude = x -> x isa Array) .=== [m4, m3, m2])
+  @test all(fcollect(m4) .=== [m1, m2, m3, m4])
+  @test all(fcollect(m4, exclude = x -> x isa Array) .=== [m2, m3, m4])
   @test all(fcollect(m4, exclude = x -> x isa Foo) .=== [m4])
 
   m1 = [1, 2, 3]
@@ -81,12 +81,12 @@ end
   m0 = NoChildren(:a, :b)
   m3 = Foo(m2, m0)
   m4 = Bar(m3)
-  @test all(fcollect(m4) .=== [m4, m3, m2, m1, m0])
+  @test all(fcollect(m4) .=== [m1, m2, m0, m3, m4])
 
   m1 = [1, 2, 3]
   m2 = [1, 2, 3]
   m3 = Foo(m1, m2)
-  @test all(fcollect(m3) .=== [m3, m1, m2])
+  @test all(fcollect(m3) .=== [m1, m2, m3])
 end
 
 struct FFoo
@@ -143,8 +143,8 @@ end
   m2 = [1, 2, 3]
   m3 = FFoo(m1, m2, (:y, ))
   m4 = FBar(m3, (:x,))
-  @test all(fcollect(m4) .=== [m4, m3, m2])
-  @test all(fcollect(m4, exclude = x -> x isa Array) .=== [m4, m3])
+  @test all(fcollect(m4) .=== [m2, m3, m4])
+  @test all(fcollect(m4, exclude = x -> x isa Array) .=== [m3, m4])
   @test all(fcollect(m4, exclude = x -> x isa FFoo) .=== [m4])
 
   m0 = NoChildren(:a, :b)
@@ -152,5 +152,5 @@ end
   m2 = FBar(m1, ())
   m3 = FFoo(m2, m0, (:x, :y,))
   m4 = FBar(m3, (:x,))
-  @test all(fcollect(m4) .=== [m4, m3, m2, m0])
+  @test all(fcollect(m4) .=== [m2, m0, m3, m4])
 end


src/functor.jl
```diff
@@ -169,7 +169,52 @@ function _default_walk(f, x)
...
-_default_walk(f, ::Nothing, ::Nothing) = nothing
+_default_walk(_, ::Nothing, ::Nothing) = nothing
+
+# Side effects only, saves a restructure
+function _foreach_walk(f, x)
+  foreach(f, children(x))
+  return x
+end
+
+### WARNING: the following is unstable internal functionality. Use at your own risk!
```
ToucheSir, line 180: I considered `_`-prefixing everything but didn't for now because these will probably become part of the public API at some point. Happy to prefix them if that's desired though.

src/functor.jl
```diff
@@ -169,7 +169,52 @@ function _default_walk(f, x)
...
+  fn::F
+  isleaf::L
+  cache::C
+  walk::W
+end
+(fld::Fold)(x) = get!(() -> fld.fn(fld.isleaf(x) ? x : fld.walk(fld, x)), fld.cache, x)
+
+# Convenience function for working with `Fold`
+function fold(f, x; isleaf = isleaf, cache = false, walk = _default_walk)
+  if cache === true
```
ToucheSir, line 210: Note that `fmap` and `fcollect` are still passing through `IdDict()` by default. The hope is that new functions either don't allow customization of the cache or use a "safer" default.

src/functor.jl
```diff
@@ -296,8 +340,8 @@ fmapstructure(f, x; kwargs...) = fmap(f, x; walk = (f, x) -> map(f, children(x))
     fcollect(x; exclude = v -> false)
 
 Traverse `x` by recursing each child of `x` as defined by [`functor`](@ref)
-and collecting the results into a flat array, ordered by a breadth-first
-traversal of `x`, respecting the iteration order of `children` calls.
+and collecting the results into a flat array, ordered by a depth-first,
+post-order traversal of `x` that respects the iteration order of `children` calls.
```
ToucheSir, line 344: This is technically breaking (as can be seen from the test changes), but it's unclear to me how many users were relying on the breadth-first iteration order. The one example I found on JuliaHub, AlphaZero, was not. It's also odd that the original worked breadth-first when pretty much all other traversals based on functors are depth-first. So RFC, and happy to back this out if we consider it too breaking.

darsnack: I think it's fine to have this breaking change.

src/functor.jl
```diff
@@ -253,11 +300,10 @@ Foo(Bar([1, 2, 3]), (40, 50, Bar(Foo(6, 7))))
 ```
 """
 function fmap(f, x; exclude = isleaf, walk = _default_walk, cache = IdDict())
-  haskey(cache, x) && return cache[x]
-  y = exclude(x) ? f(x) : walk(x -> fmap(f, x, exclude = exclude, walk = walk, cache = cache), x)
-  cache[x] = y
-
-  return y
+  return fold(x; cache, walk, isleaf = exclude) do node
+    !exclude(node) && return node
```
darsnack, line 304: Shouldn't this be
```suggestion
    exclude(node) && return node
```

darsnack (approved):
This is a great improvement!

src/functor.jl
mcabbott: Can I suggest just this?
```
exclude(node) ? f(node) : node
```
The `&& return` pattern makes more sense when there is real work done below.

src/functor.jl
```diff
@@ -169,7 +169,54 @@ function _default_walk(f, x)
...
+end
+
+### WARNING: the following is unstable internal functionality. Use at your own risk!
+# Wrapper over an IdDict which only saves values with a stable object identity
+struct Cache{K,V}
+  inner::IdDict{K,V}
+end
+Cache() = Cache(IdDict())
+
+iscachesafe(x) = !isbits(x) && ismutable(x)
```
mcabbott, line 187: Can I suggest calling this `usecache`? It's not so clear to me which way "safe" points, safe that we don't need the cache, or safe that we can use it? 

Also, why `ismutable`? This catches only the outermost wrapper, which I don't think ought to matter.

src/functor.jl
```diff
@@ -169,7 +169,54 @@ function _default_walk(f, x)
...
+# Functionally immutable and observe value semantics, but still `ismutable` and not `isbits`
+iscachesafe(::Union{String,Symbol}) = false
+# For varargs
+iscachesafe(xs::Tuple) = all(iscachesafe, xs)
+Base.get!(f, c::Cache, x) = iscachesafe(x) ? get!(f, c.inner, x) : f()
+
+# Passthrough used to disable caching (e.g. when passing `cache=false`)
+struct NoCache end
+Base.get!(f, ::NoCache, _) = f()
```
mcabbott, line 196: Style not logic, but could this be more tidily done as a `safeget!` function, rather than types on which Base's functions act? There is I think exactly one line which calls this `get!` on these types.

mcabbott:
>  e.g. fmap(f, x, xs...). This would help Optimisers.jl and any other downstream library that have rolled their own fmap variants with functor.

Can you sketch how this would work? Optimisers does some tricky things, and it often has to treat the gradient asymmetrically to the model. Is it clear that this is a good fit?

src/functor.jl
darsnack: Also, I retract my suggestion which was wrong, fine with Michael's above.

src/functor.jl
mcabbott: I think that nothing now in Optimisers.jl now cares about the order taken by `fmap`. 

Of course the two halves of `destructure` care a lot and must match. I haven't completely understood how #31 would have to change; it currently uses `fmap` for one half and something hand-written for the other.

src/functor.jl
```diff
@@ -253,11 +300,10 @@ Foo(Bar([1, 2, 3]), (40, 50, Bar(Foo(6, 7))))
...
 """
 function fmap(f, x; exclude = isleaf, walk = _default_walk, cache = IdDict())
-  haskey(cache, x) && return cache[x]
-  y = exclude(x) ? f(x) : walk(x -> fmap(f, x, exclude = exclude, walk = walk, cache = cache), x)
-  cache[x] = y
-
-  return y
+  return fold(x; cache, walk, isleaf = exclude) do node
+    !exclude(node) && return node
+    return f(node)
```
ToucheSir, line 305:
```suggestion
    exclude(node) ? f(node) : node
```

src/functor.jl
ToucheSir: Good catch, I think this was a copy-paste from fcollect.

src/functor.jl
ToucheSir: `usecache` sounds better, changed.

`ismutable` is required because this function is called at every level of the tree. So if there is a string in a (possibly deeply) nested struct that otherwise contains all bitstype values, `isbits` will fail and we will get a false positive. In fact `ismutable` may be sufficient on its own, but in the spirit of not making too many assumptions around underspecified language behaviour (namely that all mutable structs have stable objectids) I've kept it in to be safe.

src/functor.jl
```diff
@@ -169,7 +169,54 @@ function _default_walk(f, x)
...
+  inner::IdDict{K,V}
+end
+Cache() = Cache(IdDict())
+
+iscachesafe(x) = !isbits(x) && ismutable(x)
+# Functionally immutable and observe value semantics, but still `ismutable` and not `isbits`
+iscachesafe(::Union{String,Symbol}) = false
+# For varargs
+iscachesafe(xs::Tuple) = all(iscachesafe, xs)
+Base.get!(f, c::Cache, x) = iscachesafe(x) ? get!(f, c.inner, x) : f()
```
ToucheSir, line 192:
```suggestion
usecache(x) = !isbits(x) && ismutable(x)
# Functionally immutable and observe value semantics, but still `ismutable` and not `isbits`
usecache(::Union{String,Symbol}) = false
# For varargs
usecache(xs::Tuple) = all(usecache, xs)
Base.get!(f, c::Cache, x) = usecache(x) ? get!(f, c.inner, x) : f()
```

darsnack:
Haven't tested this, but this a rough sketch of the approach. Probably needs more iterations in the oven, but the more we take Functors.jl in this direction, the better. Even without Optimisers.jl, I will be able to use these improvements for pruning.

```julia
function update(tree, x, dxs...)
  function opt_walk(f, leaf, x, dxs...)
    any(isnothing, (leaf, x, dxs...)) && return first(dxs)
    stree = children(leaf)
    sx = children(x)
    sdxs = map(children, dxs...)
    foreach((_tree, _x, _dxs) -> f(_tree, _x, _dxs...), stree, sx, zip(sdxs))
  end

  fmap(apply, tree, x, dxs...; walk = opt_walk)
end
```

src/functor.jl
ToucheSir: I had exactly this in an earlier draft, but unfortunately it's a breaking change (i.e. tests fail) because the default before was to unconditionally trust IdDict/objectid/===. So value types were being deduped via the cache. If we decide to call that a bug, then I can put this suggestion in place.

Another theoretical benefit of not hard-coding `usecache` would be allowing users to customize caching behaviour on a case-by-case basis. I haven't thought of a scenario that would require such a degree of customization yet (e.g. `TiedArrays` could simply overload `usecache`), so we can cross that bridge when we come to it.

src/functor.jl
ToucheSir: The order in `fmap` has always been stable, things brings `fcollect` in line as it was the odd one out.

ToucheSir:
> Can you sketch how this would work? Optimisers does some tricky things, and it often has to treat the gradient asymmetrically to the model. Is it clear that this is a good fit?

Good questions. `fmap` will never modify the structure of the tree, so using it that way is a non-starter. I intend to rename or alias it to `mapleaves`  to clearly reflect that [^1]. `fold`, on the other hand, will traverse every level of the tree. So you can add or prune subtrees at will using it. This is overkill for most of Optimisers, but you can imagine how it might be useful for FluxML/Optimisers.jl#42 with a little tweaking.

Handling parallel traversal (i.e. zip then traverse) over asymmetric trees will require better walk functions than we currently have in Functors. I started prototyping that as part of #27, but it's going to be a challenge to get working without completely rewriting core `functor` function itself. Hence the incremental approach described up top: keep the user-facing interface as stable as possible while updating the internals. If that requires maintaining two parallel sets of functionality (one for back compat and one for the new stuff), so be it.

[^1]: What's old is new again: https://github.com/FluxML/Flux.jl/blob/e7d76b8423818c5a165e388dd3b090cc5bf42cbb/src/treelike.jl#L27. But seriously, removing jargon from functors is a good thing, especially when some of it (e.g. `fmap`) is not strictly correct.

mcabbott:
Ok. It might be worth trying to write FluxML/Optimisers.jl#42 using this as a logic stress-test? Although need not hold up this PR I guess. (I am not happy yet with how to handle transposed tied arrays, there, for which the gradient might not have .parent, but might be a thunk or a Broadcasted....)

Longer term I think we should be open to just scrapping Functors and starting over. The entire library fits on a page, really, it's just a matter of choosing logic which matches the problems we want to solve. And a clean start (or two or three) might be much easier than contortions to modify it while keeping backward compat.

darsnack:
The major breakages will be w.r.t. Flux's uses of Functors. There are probably few users directly using Functors outside of `@functor`. So, if we need to scrap one day, then the contortions can always live in Flux.jl for a deprecation cycle.

src/functor.jl
mcabbott: Right, my suggestion still leaves `usecache` as an overloadable function. It's not intended as a change of logic, just more compact & fewer types?

Testing `isbits` etc. seems at all seems like a potentially breaking change, as it changes how many times `f` will be called (e.g. on numbers in a tuple). Maybe it depends how firmly you believe `f` must be pure... `fcollect` is not I think.

mcabbott:
We could contemplate having Flux provide an `@functor` which for now calls this one, to ease transitions.

ToucheSir:
The entire library fitting on a page is both part of the allure and the fundamental problem with Functors. It makes the problem look so easy, whereas as we've discovered it is anything but.

> Longer term I think we should be open to just scrapping Functors and starting over.

This could be done today by moving to an alternative like https://github.com/chengchingwen/StructWalk.jl. `fold` is basically `postwalk` + cache and overriding of `isleaf`, after all. IMO `WalkStyle` is a cleaner and more idiomatic paradigm than custom walk functions too. In any case, the missing pieces for any contender (including Functors itself) are:
1. Walking over similarly structured trees, `walk(f, x, xs...)`
2. Walking over multiple subtrees with different branches, `walk(f, (a=1, b=2), (b=3, c=4))`
3. Functionality to traverse multiple trees and return a single tree, `fmap(f, x, xs...) -> tree`
3. Functionality to traverse multiple trees and return multiple trees, `fmultimap(f, x, xs...) -> tree, trees...`. This could also be done as an `unzip`-like post-processing step, see [`jax.tree_util.tree_transpose`](https://jax.readthedocs.io/en/latest/_autosummary/jax.tree_util.tree_transpose.html?highlight=transpose#jax.tree_util.tree_transpose).

cc @chengchingwen for his thoughts on this.

> We could contemplate having Flux provide an `@functor` which for now calls this one, to ease transitions.

JuliaHub tells me that a number of libraries rely on Functors but not Flux, so it would help a bit but not completely solve the migration issue.

chengchingwen:
@ToucheSir Do you have some small test cases in mind for those missing pieces? It's not obvious to me how they should work in general. For example, what would be considered to be similarly structured: Is `1//2` similar to `(a=1,b=2)`? what about `(a, b)` and `(b=3, c=4)` (they can both be viewed as `NTuple{2}`)? Or maybe we want an interface to define what is similar and how should they aligned? I used to tackle a similar problem in [loading weights](https://github.com/chengchingwen/Transformers.jl/blob/432eb9b7bfcdbe21b49fcbdc933b7ff9cc8aa842/src/huggingface/models/models.jl#L42-L55) from `state_dict` by using the field names (i.e. `hasfield`/`getfield`/`keys`/...) and `functor`, and `StructWalk.jl` was originally built for rewriting that part of code. OTOH, what is the different between "walking over subtrees" (1./2.) and "traverse multiple trees" (3./4.)? Returning multiple trees (4.) is the same as returning a tuple of trees, which is a single tree (3.). And if we are traversing multiple trees that are not similar / cannot be aligned (3. but not 1./2.), what would the traversal algorithm be? Priority first search with custom priority?

src/functor.jl
```diff
@@ -169,7 +169,54 @@ function _default_walk(f, x)
...
   re(map(f, func))
 end
-_default_walk(f, ::Nothing, ::Nothing) = nothing
+_default_walk(_, ::Nothing, ::Nothing) = nothing
+
+# Side effects only, saves a restructure
+function _foreach_walk(f, x)
+  foreach(f, children(x))
+  return x
+end
```
mcabbott, line 178: Can I suggest making:
```
function _structural_walk(f, x)
  func, re = functor(x)
  map(f, func)
end
```
This can be used by `fmapstructure` and `fcollect`, and an `fmapnothing` could just throw away the return.

Are there efficiency reasons to have a 3rd walk with explicit `foreach`? I guess if you have a non-leaf array, this will save allocating the result.

ToucheSir:
@chengchingwen those are all good examples, and honestly I'm not sure either. One idea I had is that structural similarity would be determined by the LHS. So `(1, 2), (a=1,b=2), ...` uses the tuple path, while `(a=1,b=2), (1, 2), ...` uses the namedtuple path. Arrays will be similar to tuples while structs will act like namedtuples (I don't want to think about Dicts...).

>  And if we are traversing multiple trees that are not similar / cannot be aligned (3. but not 1./2.), what would the traversal algorithm be? Priority first search with custom priority?

That too I'm not sure on. Do we error/warn about them, if at all? My thought was to take the intersection of all properties for keyed types and walk over only those. Some thinking out loud with pseudocode:
```julia
function walk(f, nt, nts...) # single return, f(xs...) = y
  common_props = intersect(propertynames(nt), map(propertynames, nts...)...)
  patch = map(f, nt[common_props], map(t -> t[common_props])...)
  return merge(nt, patch)
end

function walk(f, nt, nts...) # multiple returns, f(xs...) = ys..., length(ys) == length(xs)
  common_props = intersect(propertynames(nt), map(propertynames, nts...)...)
  patch, patches... = map(f, nt[common_props], map(t -> t[common_props])...)
  return (merge(nt, patch), map(merge, nts, patches)...)
end
```
Maybe that's too surprising?

Linearly-indexed types should probably just fail on a length mismatch? I see the 2-arg case as an instance of the multi-arg case, but perhaps that's the wrong way to look at it.

> Returning multiple trees (4.) is the same as returning a tuple of trees, which is a single tree (3.)

Yes, the question is how to create a tuple of trees given a tuple of trees and a user-specified varargs function. AIUI applying the function at each node will result in a tree of tuples instead, so the library would need to know how to unwrap those tuples either during or after traversal.

chengchingwen:
@ToucheSir  
>  One idea I had is that structural similarity would be determined by the LHS

This might be problematic since the similar operator is asymmetric, and I'm not sure we always want to align the children just according to the first argument (or just the intersect of property names as in the pseudocode). There are cases we would align children with different name or even different depths to ignore some wrapper. 

All in all, I think we need to clarify what the data do we really need during the walk/transformation. In the "walking over subtrees" paradigm, we want the "zip"ed data in each tree like running several BFS in parallel. OTOH, In the "traverse multiple trees" paradigm, we might want different part of data. So IMPO no matter it's a single/multiple input tree single/multiple output tree function, the real question is in what order do we want to access those subtrees?

src/functor.jl
mcabbott: Maybe I start to agree, and having just `walk = (f, x) -> map(f, children(x))` (as now) and `walk = (f, x) -> foreach(f, children(x))` around is better than giving these names. So maybe I retract the suggestion.

src/functor.jl
ToucheSir: Checking for `isbits` and `ismutable` is a hedge against an impure `f`. IMO users would be more surprised if `fmap(println, (1,2,3,1,2,3))` printed `1\n2\n3\n` as it currently does.

mcabbott: Yes. I'm in favour of some such check. But perhaps this can be factored out, and we can discuss it this one has the right rules. (Should it be any or all?) This seems largely orthogonal.

ToucheSir:
> This might be problematic since the similar operator is asymmetric, and I'm not sure we always want to align the children just according to the first argument (or just the intersect of property names as in the pseudocode). There are cases we would align children with different name or even different depths to ignore some wrapper.

That's true, is there a way to customize this alignment behaviour? To be clear, I only think basing the output structure off the leftmost tree makes sense when a single tree is returned from the multi-tree walk, as it aligns somewhat with how `map` works on collections now. Having it be pluggable with a sane default (e.g. force exact structural similarity as [`liftA2`](https://hackage.haskell.org/package/base-4.16.0.0/docs/Control-Applicative.html#v:liftA2) does) would be amazing.

> All in all, I think we need to clarify what the data do we really need during the walk/transformation. In the "walking over subtrees" paradigm, we want the "zip"ed data in each tree like running several BFS in parallel. OTOH, In the "traverse multiple trees" paradigm, we might want different part of data. So IMPO no matter it's a single/multiple input tree single/multiple output tree function, the real question is in what order do we want to access those subtrees?

I think "walking over multiple subtrees" was bad wording on my part. A better term would've been "dissimilar trees" to mirror the 1st point about similarly-structured trees. Presumably that narrows down the design space dramatically, as I can't think of a use-case in e.g. Optimisers.jl that wouldn't work with either `prewalk` or `postwalk`.

chengchingwen:
Some side notes about `StructWalk.jl` is that it has a interface for define custom `WalkStyle`, so we can have different kinds of children for different types according to the style. This is something that I think is missing in the current implementation of `Functors.jl`, as we are all based on `functor` to define the tree-equivalent `map` (`trainable`/`gpu`/`f32`/...), but those function might actually need different sets of child node. For example, not all array fields are `trainable`, but some should still be `gpu`-able. So the `WalkStyle` interface provide a way do modified what is being treated as a child node. Besides, `walk` are more general than `fmap` as they not only apply function on the leaf nodes, it also apply on the non-leaf nodes. I'm not sure if we really need this for the optimiser case, but should totally be ignored if not needed.

ToucheSir:
> One thing it wants is the tree equivalent of `map(f, pairs(x))`, with cache=false, over only trainable nodes.

`map(f, pairs(x))` over trainable nodes is doable with a custom walk function on master, I believe. The logic would be similar to `Optimisers._trainable`, but with a callback. One could also envision another walk that does replace non-trainable fields with nothing. In fact I think you could MacGyver one with the current functionality in Optimisers: 
```julia
# _trainable(ch::NamedTuple, tr::NamedTuple) = merge(map(_ -> nothing, ch), tr) ...

function _trainable_fillnothing_walk(f, x)
  func, re = functor(x)
  tfunc = trainable(x)
  return re(_trainable(func, map(f, tfunc)))
end
```
`cache=false` is part of this PR, but could be accomplished on master with a dummy dict type with dummy `setindex!` and `haskey` methods.

> As a concrete example, can this be written neatly with pre/post-walk? Or without, is it made easier / neater by this PR?

`fmap` is already doing an implicit post-order traversal, so I'd say so. This PR makes that implicit part explicit by pulling it into a function. What `fold` buys you over plain `fmap` is that it visits non-leaf nodes too. I think most of Optimisers could get away without such a thing (via packing logic into walk functions, assuming we get multi-tree walk functions), but it has the potential to make things neater.

To Peter's latest point, all this is why I became interested in StructWalk.jl. `fold` is more or less a convergently evolved `StructWalk.postwalk`, but we have no equivalent of `prewalk` in Functors at present. IMO `WalkStyle` is far nicer than custom walk functions (which were a bit of a hack from the beginning, we basically exposed part of the internals). The only thing missing is the ability to walk multiple trees (regardless of semantics), but perhaps that discussion would be better had on the StructWalk issue tracker. Thoughts?

src/functor.jl
ToucheSir: So to clarify, yea to calling the current behaviour a bug and unconditionally calling `usecache` if it helps get rid of the structs?

PR #33, opened 2022-01-31
Functor `Transpose` et. al.

mcabbott:
Part of #28. Better implementation than that in #31. 

There is some chance this will break Flux's current Params thing horribly, I should check.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -3,6 +3,10 @@
 
 functor(::Type{<:Base.ComposedFunction}, x) = (outer = x.outer, inner = x.inner), y -> Base.ComposedFunction(y.outer, y.inner)
 
+###
+### Array wrappers
+###
+
 using LinearAlgebra
 # The reason for these is to let W and W' be seen as tied weights in Flux models.
 # Can't treat ReshapedArray very well, as its type doesn't include enough details for reconstruction.
@@ -19,6 +23,9 @@ _transpose(x) = transpose(x)
 _transpose(x::NamedTuple{(:parent,)}) = x.parent
 _transpose(bc::Broadcast.Broadcasted{S}) where S = Broadcast.Broadcasted{S}(_conjugate(bc.f, transpose), _transpose.(bc.args))
 
+_conjugate(f::F, ::typeof(identity)) where F = f
+_conjugate(f::F, op::Union{typeof(transpose), typeof(adjoint)}) where F = (xs...,) -> op(f(op.(xs)...))
+
 function functor(::Type{<:PermutedDimsArray{T,N,perm,iperm}}, x) where {T,N,perm,iperm}
   (parent = _PermutedDimsArray(x, iperm),), y -> PermutedDimsArray(only(y), perm)
 end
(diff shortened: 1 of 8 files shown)


src/base.jl
```diff
@@ -1 +1,34 @@
+
 @functor Base.RefValue
+
+functor(::Type{<:Base.ComposedFunction}, x) = (outer = x.outer, inner = x.inner), y -> Base.ComposedFunction(y.outer, y.inner)
+
+using LinearAlgebra
+# The reason for these is to let W and W' be seen as tied weights in Flux models.
+# Can't treat ReshapedArray very well, as its type doesn't include enough details for reconstruction.
+
+functor(::Type{<:Adjoint}, x) = (parent = _adjoint(x),), y -> adjoint(only(y))
```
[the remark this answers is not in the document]
mcabbott: Yes. The point of `_adjoint` is the it digests both, it will `adjoint(dx::Matrix)` or unwrap `dx::Adjoint` (both natural), and deal with `dx::NamedTuple`. I'm not honestly 100% sure why the structural ones show up, there is some chance that could always be avoided.

But the fact that `functor` can be used on `dx` with a type from `x` is, I now think, the key feature to make it useful for Optimisers.

src/base.jl
```diff
@@ -1 +1,34 @@
...
+
+_adjoint(x) = adjoint(x)  # _adjoint is the inverse, and also understands more types:
+_adjoint(x::NamedTuple{(:parent,)}) = x.parent  # "structural" gradient, and lazy broadcast used by Optimisers:
+_adjoint(bc::Broadcast.Broadcasted{S}) where S = Broadcast.Broadcasted{S}(_conjugate(bc.f, adjoint), _adjoint.(bc.args))
+
+functor(::Type{<:Transpose}, x) = (parent = _transpose(x),), y -> transpose(only(y))
+
+_transpose(x) = transpose(x)
+_transpose(x::NamedTuple{(:parent,)}) = x.parent
+_transpose(bc::Broadcast.Broadcasted{S}) where S = Broadcast.Broadcasted{S}(_conjugate(bc.f, transpose), _transpose.(bc.args))
```
mcabbott, line 20: Some chance this Broadcasted stuff is too Optimiser-specific. But piracy to put it there? The less elaborate version is to give up and `materialize(bc)` like PermutedDimsArray here.

Ideally, I think this would actually move to ChainRulesCore, as some kind of BroadcastThunk object. But... maybe later.

ToucheSir: Why would it be piracy to include in Optimisers over Functors? One way to avoid it could be to pull the thread a little harder and bring anything connected to it over to Optimisers as well. Depends how much we see Functors as a minimal "core" library vs something that should provide a reasonable OOTB experience.

mcabbott: I literally mean that `Functors._transpose(bc::Broadcast....)` is piracy.

mcabbott:
Flux's tests pass with this. 

It does not in fact seem to enable Flux to see M and `transpose(M)` as sharing weights, perhaps that requires some other change:
```julia
julia> mat = [1 2; 3 4.0];

julia> fmapstructure(println, (mat, mat'))  # prints once
[1.0 2.0; 3.0 4.0]
(nothing, (parent = nothing,))

julia> Flux.params((mat, mat'))  # still sees two arrays
Params([[1.0 2.0; 3.0 4.0], [1.0 3.0; 2.0 4.0]])

(jl_9EHJ8B) pkg> st
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_9EHJ8B/Project.toml`
  [587475ba] Flux v0.13.3 `~/.julia/dev/Flux`
  [d9f16b24] Functors v0.3.0 `~/.julia/dev/Functors`
  [3bd65402] Optimisers v0.2.6 `~/.julia/dev/Optimisers`
```
But it will allow https://github.com/FluxML/Optimisers.jl/pull/42 to see such weight sharing.

ToucheSir (approved):
For Flux, https://github.com/FluxML/Flux.jl/blob/674527e0403c8030baf5f4f15835aff2caddfbc6/src/functor.jl#L39 would need changing to exclude wrapped array types.

mcabbott:
Great, thanks for looking. That should probably test `isleaf`.

PR #36, opened 2022-02-05
Return empty NamedTuple in `@functor` instead of empty Tuple

ToucheSir:
https://github.com/FluxML/Flux.jl/pull/1862 made me realize that this was inconsistent.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -16,7 +16,7 @@ function makefunctor(m::Module, T, fs = fieldnames(T))
   escfs = [:($f = getfield(x, $(QuoteNode(f)))) for f in fs]
 
   @eval m begin
-    $Functors.functor(::Type{<:$T}, x) = ($(escfs...),), y -> $T($(escargs...))
+    $Functors.functor(::Type{<:$T}, x) = (;$(escfs...)), y -> $T($(escargs...))
   end
 end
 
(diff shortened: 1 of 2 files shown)


mcabbott:
Maybe a test?

ToucheSir:
Need to think about the ramifications this has on `isleaf` a bit first, since that checks `=== ()`. I think this is more correct, but it's also breaking unless `isleaf` is changed too.

ToucheSir:
Latest commit mainly adds some tests for the existing changes. After some thought, not having `isleaf` check for `NamedTuple{{}, Tuple{}}` does make some sense. This allows one to distinguish between an "empty" node and a leaf when performing traversals, thus making `exclude` functions more precise. One could argue `nothing` or a custom singleton struct might be a better sentinel value for this, but we'd want better downstream testing before trying that approach.

As a bonus, this brings normal `@functor` in line with `@flexiblefunctor`, which has been generating a NamedTuple this whole time.

darsnack (approved):

mcabbott:
To check I understand right, what this will change is things like this:
```julia
julia> using Functors: @functor, functor

julia> functor(nothing)
((), Returns{Nothing}(nothing))

julia> @functor Nothing  # struct with no fields

julia> functor(nothing)  # this will change to (;), ...
((), var"#7#8"())
```

ToucheSir:
That's correct. I don't believe we have any field-less structs in FluxML, and while norm layers (except `LayerNorm`) can return `(;)` if they're not affine that is for `trainable`, not `functor`.

PR #37, opened 2022-02-06
Make `fmap(f, x, y)` useful

mcabbott:
This alters `fmap(f, x, y)` to more closely agree with  `fmap(f, x)`, in particular to take kwywords `exclude, walk`.

And it inserts `functor(typeof(x), y)`, so that if not all fields of `x` are children, the same fields of `y` are used. The first object `x` is the primary one, whose objects are stored in the cache. But there is no longer any assumption that the 2nd is a gradient, or that Nothing is special.

Third, it adds a `prune` keyword, which allows you to instead use a constant for all repeated leaves, rather than the original. ~~Probably the same keyword should be added to `fmap(f, x)`.~~

What motivates this is an attempt to write `destructure` using `fmap(x, y)`, https://github.com/FluxML/Optimisers.jl/pull/54. Where in fact `y` is a tree of offsets; and the walk which includes the gradients doesn't really want the cache, since it needs to look at all leaves.

These changes don't break any existing tests. But should of course have better tests, and docs.


ToucheSir:
Looking at the state of kwargs for `fmap`, I'm of the opinion that we should consider moving the responsibility for caching + pruning out of the function. `fmap` and co would then be natively differentiable, and there would be fewer params to thread through the call stack. That's not to say we should remove caching from Functors altogther. Instead, I think they could be handled by a type that wraps the callback function and encapsulates any caching + pruning related state. This would also open up the opportunity to customize caching behaviour on a case by case basis. Users who are fine with a plain tree traversal sans duplicate checking experience no additional overhead, while those who want just a subset of this functionality can pay for what they use.

mcabbott:
I can believe there might be cleaner ways to do things. But for now I propose baby steps, in part to feel out what's needed --- accidentally choosing a cleaner design which prohibited `prune` would make `destructure` harder.

src/functor.jl
```diff
@@ -43,12 +43,11 @@ function _default_walk(f, x)
   re(map(f, func))
 end
 
-function fmap(f, x; exclude = isleaf, walk = _default_walk, cache = IdDict())
-  haskey(cache, x) && return cache[x]
-  y = exclude(x) ? f(x) : walk(x -> fmap(f, x, exclude = exclude, walk = walk, cache = cache), x)
-  cache[x] = y
+const INIT = Base._InitialValue()  # sentinel value for keyword not supplied
```
darsnack, line 46: Maybe less opaque to just define `struct NoPrune end`?

mcabbott: Can do. Are you OK with the interface being that the *absence* of this keyword is how you turn this off? That's how `init` often works; and here, the values you might use for "don't prune" like `false` or `nothing` are all reasonable choices for the value to use during pruning.

darsnack: The code is weird to me, but the API makes sense. I'm happy.

src/functor.jl
```diff
@@ -43,12 +43,11 @@ function _default_walk(f, x)
...
-function fmap(f, x; exclude = isleaf, walk = _default_walk, cache = IdDict())
-  haskey(cache, x) && return cache[x]
-  y = exclude(x) ? f(x) : walk(x -> fmap(f, x, exclude = exclude, walk = walk, cache = cache), x)
-  cache[x] = y
+struct NoKeyword end
 
-  return y
+function fmap(f, x; exclude = isleaf, walk = _default_walk, cache = IdDict(), prune = NoKeyword())
+  haskey(cache, x) && return prune isa NoKeyword ? cache[x] : prune
+  cache[x] = exclude(x) ? f(x) : walk(x -> fmap(f, x; exclude, walk, cache, prune), x)
```
mcabbott, line 50: To use keywords like this, this package cannot claim to support Julia 1.0. At present it is only tested on 1.5+. Maybe we should just move to 1.6?

ToucheSir: We already have on CI, might as well make it official.

ToucheSir:
> I can believe there might be cleaner ways to do things. But for now I propose baby steps, in part to feel out what's needed --- accidentally choosing a cleaner design which prohibited `prune` would make `destructure` harder.

This is actually my worry with adding yet another kwarg. Removing it requires a breaking change and deprecation cycle, which sucks. Let me cook up something tonight for the wrapper version, because I feel like the keyword approach is probably the less flexible of the two.

src/functor.jl
ToucheSir: I'm not sure what repo I was looking at, because I just checked back and it's 1.5 still. Do we need to cut a breaking release for minimum version bumps like this again?

ToucheSir:
Ok, after trying out some things I'm generally ok with this given a migration path like the following:
1. Create a separate function that doesn't rely on (as many kwargs) and give it a better name.
2. Switch downstream over to using the new function
3. Deprecate and remove `fmap`. I'd love to use the name for something more appropriate if possible. For example, what we're currently calling "walk" is really what FP people use `fmap` for, not nested (heterogeneous) structure traversal.

src/functor.jl
mcabbott: It's 1.0 here: https://github.com/FluxML/Functors.jl/blob/master/Project.toml#L7

mcabbott: And, people say not to do this on a patch release, because it closes the door on a bugfix-for-1.0 release. Do we care? 

We may also need a breaking release for #33, and to make the cache not used on isbits arguments. We could gang these together.

mcabbott:
Ok. I'm a little worried that after this re-design I won't understand this library at all! 

But for now I think it would be great to get the things which work today merged and in use. Rather than wait for a cleaner design and then wait for someone to have time to re-write Optimisers to use it... What's the plan here, is going to Julia 1.6 on v0.2.8 ok? This is not otherwise breaking, and I am happy to walk away from 1.0 forever. Or should this be made to work on 1.0, for now, until enough other things are ready to justify 0.3?

ToucheSir:
You've already seen 3/4ths of the "re-design" in #32, so I wouldn't be too worried about that :wink:.

On the lack of compatibility, what exactly doesn't work on 1.0? I didn't really understand why the existing usage of keyword args was fine but the additions in this PR aren't.

mcabbott:
[de27d33](https://github.com/FluxML/Functors.jl/pull/37/commits/de27d3307497513721f6b24a32160b1d9d0a14a5) might let it work on 1.0. Tests fail as it is unable to get a working version of Zygote.

mcabbott:
Good to go? 

No tagged version has `fmap(f, xs...)` so this can't break anything.

ToucheSir (approved):

PR #42, opened 2022-06-01
Fix missing docs and cross-references

Saransh-cpp:
Some minor changes in the documentation. These changes would help us in fixing the broken links and references in `Flux`'s documentation. 

* make the docs of `@functor` render in the manual
* fix cross-references within the docs
* update `.gitignore`

**Note**: The current docstring of `@functor` does not render, probably because of the `var""` keyword.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .DS_Store
 /dev/
 Manifest.toml
+build
+.vscode
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -30,47 +30,6 @@ function functorm(T, fs = nothing)
   :(makefunctor(@__MODULE__, $(esc(T)), $(fs...)))
 end
 
-"""
-    @functor T
-    @functor T (x,)
-
-Adds methods to [`functor`](@ref) allowing recursion into objects of type `T`,
-and reconstruction. Assumes that `T` has a constructor accepting all of its fields,
-which is true unless you have provided an inner constructor which does not.
-
-By default all fields of `T` are considered [`children`](@ref); 
-this can be restricted be restructed by providing a tuple of field names.
-
-# Examples
-```jldoctest
-julia> struct Foo; x; y; end
-
-julia> @functor Foo
-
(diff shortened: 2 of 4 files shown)


src/functor.jl
```diff
@@ -30,6 +30,47 @@ function functorm(T, fs = nothing)
...
+julia> ch2
+(a = 10, c = 30)
+
+julia> re3(("ten", "thirty"))
+TwoThirds("ten", 20, "thirty")
+
+julia> fmap(x -> 10x, TwoThirds(Foo(1,2), Foo(3,4), 56))
+TwoThirds(Foo(10, 20), Foo(3, 4), 560)
+```
+"""
```
ToucheSir, line 73: Duplicating the docstring here is not a satisfactory solution. It would be better to figure out why the one in the main package file doesn't work and fix that link instead. Perhaps `@doc` could be useful here.

.gitignore
```diff
@@ -1,3 +1,5 @@
 .DS_Store
 /dev/
 Manifest.toml
+build
```
ToucheSir, line 4: Where does this directory come from? AFAICT it should never be generated.

ToucheSir (changes requested):

.gitignore
Saransh-cpp: This is the generated documentation's build directory, containing the HTML pages.

src/functor.jl
```diff
@@ -30,6 +30,47 @@ function functorm(T, fs = nothing)
...
 macro functor(args...)
   functorm(args...)
 end
```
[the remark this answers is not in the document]
Saransh-cpp: Using `@doc` works here!

Saransh-cpp:
The last commit turns off the doctests while building the docs, which they currently are (indirectly) due to the pinned version of julia. The doctests will still run in the CI where it is specified. (similar to - https://github.com/FluxML/Flux.jl/issues/1914)

ToucheSir (approved):
Thanks @Saransh-cpp !

Issue #41, opened 2022-05-04
Taggable functors


darsnack:
In Flux, we have `trainable` to designate a subset of leaves as nodes to walk when updating parameters for training. In FluxPrune.jl, I defined `pruneable` to designate a subset of leaves for pruning (note that these cannot be the same as the `trainable` nodes).

Right now this creates an unfortunate circumstance as discussed in https://github.com/FluxML/Flux.jl/issues/1946. Users need to `@functor` their types, remember to define `trainable` if necessary. Potentially, to use FluxPrune.jl, they might want to remember to define `pruneable`. On the developer side of things, we can use the `walk` keyword of `fmap` to walk the differently labeled leaf nodes. But this usually requires defining a separate walk function based on the subset that you are hoping to target.

An alternative would be to build this information directly into what `@functor` defines. Right now, each child of a functor has a name and a value. I propose adding "tags" which would be a tuple of symbols. Then we could do something like
```julia
@functor Conv trainable=(weight, bias) pruneable=(weight,)
```
Ideally, this mechanism should be dynamic, meaning that if Flux.jl already defines the trainable leaves of a type, then another package like FluxPrune.jl should be able to add a pruneable tag on top of that.

My hope is that we make it easier on users by only having one line for making your type Flux-compatible. And we make it easier on developers by making it easy to filter nodes when walking by tag. I haven't spent a lot of time on the implementation aspect, but I just wanted to float the notion of tags first and get some feedback.

ToucheSir:
Definitely thought about this before. The part I got stuck on was where/how to store this tag metadata. Do you have any proposals there?

darsnack:
I was thinking that `functor` should be broken up. Instead we can have `children(x::T)`, `rebuilder(::Type{T})`, and `tags(::Type{T})`. Of course, the convenience macro, `@functor`, would define all three. I haven't thought too deeply about a simple convenience syntax for the macro that would support the initial declaration + adding more tags.

`children` would return the named tuple that is returned by `functor`, and `rebuilder` would return the function that puts the struct back together. `tags` would return a named tuple similar to `children` but instead of actual values for each key, it stores a tuple of symbols corresponding to the tags (defaulting to empty).

mcabbott:
Maybe `@functor T` defines `functor(::Type{T}, x, ::Val)`, and `@functor T trainable=(x,y)` defines that and also `functor(::Type{T}, x, ::Val{:trainable})`.

[...]

PR #43, opened 2022-10-06
Separate walks out from `fmap` and add #39 to `fcollect`

darsnack:
This is a non-breaking step towards separating all the keywords out of `fmap`. It was a side-effect of working on #41, so I separated the changes out to be evaluated independently. For #41, I am considering a more significant, but orthogonal, set of changes. Do we like this refactor?

~~Actually, this does seem to be breaking, since the `walk` keyword can no longer be an arbitrary function. I only introduced `AbstractWalk` to get around the method ambiguity error for `fmap`. Don't see another way around it.~~


darsnack:
It seems previews require using Netlify? @Saransh-cpp do you have any experience here that avoids Netlify?

ToucheSir:
Looks solid. Test fixes aside, perhaps we could wrap up #39 quickly and reuse the logic from that PR in `CachedWalk`/`CollectWalk`? It would be nice to finally decouple `fmap` and caching :)

mcabbott:
Have not read this closely, but if I understand right on first look, it doesn't significantly change the logic of basic case?

Am curious whether you see such changes allowing all the weird cases in Optimisers.jl to be better abstracted. 

Re breaking, I think you're saying that the method `fmap(walk::AbstractWalk, f, x, ys...)` wants a type to distinguish from `fmap(f, x, ys...)`. But is there any need to make these both the same function? The first could be `_fmap`, maybe that's clearer anyway, and then it could allow `walk::Any`. 

(But breaking isn't a huge deal, according to https://juliahub.com/ui/Packages/Functors/lER0H/0.3.0?page=2 only 5 direct deps, although that misses Flux somehow.)

ToucheSir:
> But is there any need to make these both the same function? The first could be `_fmap`, maybe that's clearer anyway, and then it could allow `walk::Any`.

Thinking about the deprecation cycle for `cache=<not Nothing>` in `fmap`, I think it would make sense for this to be part of the public API. Could walks wrap the callback function (i.e. `recurse`)? Then both overloads would have the same arity.

darsnack:
> it doesn't significantly change the logic of basic case?

Yeah that was the goal here. I wanted to refactor the code as part of logical changes, so I pulled the refactoring out separately. Here the refactor is only a cleaner way of implementing the original logic.

> Re breaking, I think you're saying that the method fmap(walk::AbstractWalk, f, x, ys...) wants a type to distinguish from fmap(f, x, ys...). But is there any need to make these both the same function? The first could be _fmap, maybe that's clearer anyway, and then it could allow walk::Any.

I actually want to move towards `fmap(walk, f, x, ys...)` and ditch the keywords in the long run. It's definitely longer to write it out this way, but I think the logic is clearer to understand than explaining how a bunch of keywords interact in the docstring of `fmap`. I only left `fmap(f, x, ys...; kwargs...)` for backwards compatibility.

I think it should be possible to maintain backwards compatibility and support both modes. We can have the `walk` keyword get wrapped in a `AnonymousWalk` struct (internal). Or we can just choose to be breaking. I think there are fewer uses of custom `walk` functions than dependents of Functors.jl. And porting existing ones is fairly simple since the interface is the same.

> Could walks wrap the callback function (i.e. recurse)?

Part of what makes the walks composable is that `DefaultWalk` does not need to know that it's recursion calls back into `CachedWalk`. You could have an outer wrapper that holds a walk and a recursion function together. I think then we start getting to #32 which could be a good thing. I'd be willing to take this PR in that direction if it is desired. The initial goal here was to be as close to the original code as reasonable.

> Am curious whether you see such changes allowing all the weird cases in Optimisers.jl to be better abstracted.

Only in part. The walk function here is a specification for starting at Node A in the tree and recursing into its children. I am thinking about it as a rule. You could certainly replace `recurse` with two functions—`ascend` and `descend`—that allows the walk function to choose to walk back up the tree. In this way, I think you could specify the rule for finding shared nodes which Optimisers.jl wants. I don't know how much clearer or more efficient that is compared walking the tree multiple times.

One thing that already works is writing a walk that holds external information (e.g. `CachedWalk`). Then the gradient cache from https://github.com/FluxML/Optimisers.jl/pull/106 can be written more cleanly as a walk struct with a cache.

mcabbott:
Re breaking, I marked #39 as being v0.4, although I'm not entirely sure we must.

Maybe 0.4 should contain both changes? No great rush for #39. And perhaps it's worth adding the `AnonymousWalk` idea, maybe it prints a depwarn... then 0.4 will in practice probably not break anyone's code.

darsnack:
> And perhaps it's worth adding the AnonymousWalk idea, maybe it prints a depwarn... then 0.4 will in practice probably not break anyone's code.

This is now in there. I think we can release this and #39 together in 0.4 if that's desired. Otherwise, if you would rather wait for #39, that's fine too.

> Also, thoughts on whether https://github.com/FluxML/Flux.jl/pull/1932 ought to wait for any changes here?

Not on this, but my next planned change will make `functor` opt-out instead of opt-in (via ConstructionBase defaults). This means that `@layer` will be restricted to just pretty printing, so a more accurate name would be nice. It also makes `@layer` optional then instead of required.

darsnack:
Okay this now has #39 changes as well. Should be ready for another review.

src/maps.jl
```diff
@@ -0,0 +1,18 @@
...
+
+function fmap(f, x, ys...; exclude = isleaf,
+                           walk = DefaultWalk(),
+                           cache = IdDict(),
+                           prune = NoKeyword())
+  _walk = if isnothing(cache)
+    ExcludeWalk(AnonymousWalk(walk), f, exclude)
+  else
+    CachedWalk(ExcludeWalk(AnonymousWalk(walk), f, exclude), prune, cache)
+  end
```
ToucheSir, line 11:
```suggestion
  _walk = ExcludeWalk(AnonymousWalk(walk), f, exclude)
  if isnothing(cache)
    _walk = CachedWalk(_walk, prune, cache)
  end
```

src/walks.jl
```diff
@@ -0,0 +1,159 @@
...
+abstract type AbstractWalk end
+
+"""
+    AnonymousWalk(walk_fn)
+
+Wrap a `walk_fn` so that `AnonymousWalk(walk_fn) isa AbstractWalk`.
+This type only exists for backwards compatability and should be directly used.
+Attempting to wrap an existing `AbstractWalk` is a no-op (i.e. it is not wrapped).
+"""
+struct AnonymousWalk{F} <: AbstractWalk
```
ToucheSir, line 28:
```suggestion
struct AnonymousWalk{F<:AbstractWalk} <: AbstractWalk
```
Is there any reason not to constrain these type parameters, at least for now?

src/walks.jl
```diff
@@ -0,0 +1,159 @@
...
+  cache::Base.IdSet{Any}
+  output::Vector{Any}
+end
+CollectWalk() = CollectWalk(Base.IdSet(), Any[])
+
+# note: we don't have an `OrderedIdSet`, so we use an `IdSet` for the cache
+# (to ensure we get exactly 1 copy of each distinct array), and a usual `Vector`
+# for the results, to preserve traversal order (important downstream!).
+function (walk::CollectWalk)(recurse, x)
+  x in walk.cache && return walk.output
```
ToucheSir, line 152: Given this will be in 0.4, could we take the opportunity to clean up the caching behaviour? I was thinking to share logic with `usecache` so we're not accidentally identifying isbits types like SArrays as already visited.

src/walks.jl
darsnack: The reason this type exists is to wrap `F` that isn't a `AbstractWalk` but still satisfies the `AbstractWalk` interface.

src/walks.jl
darsnack: Since #39 didn't touch this, I didn't either, but I can make this apply the same caching rules.

darsnack:
Okay should be good?

ToucheSir (approved):

PR #47, opened 2022-11-14
Add more base types

### LINKED ISSUE #28: Add more Base types

mcabbott:
Closes #28, intended for 0.4 release

So far aims to add all of Base.Iterators, and Fix1, Fix2, Pair. Are there other obvious things to add?

Needs tests.

### PR Checklist

- [x] Tests are added
- [ ] Documentation, if applicable

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -49,7 +49,7 @@ _PermutedDimsArray(bc::Broadcast.Broadcasted, iperm) = _PermutedDimsArray(Broadc
 ###
 
 @functor Iterators.Accumulate
-@functor Iterators.Count
+# Count
 @functor Iterators.Cycle
 @functor Iterators.Drop
 @functor Iterators.DropWhile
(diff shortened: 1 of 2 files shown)


CarloLucibello:
I still don't understand why we shouldn't switch to an opt out semantics

mcabbott:
I would favour doing that, eventually. It's probably got to be done with a breaking release of Flux.

This PR moves us a little closer to traversing everything. Fix2 in particular came up in https://github.com/FluxML/Flux.jl/issues/2107

darsnack:
Same, I am in favor (and in fact have a branch doing something like this), but it will require many more tests than we currently have. It would be a candidate for v0.5, but it would be nice to get v0.4 out soon.

CarloLucibello (approved):

PR #48, opened 2022-11-14
Change `getproperty` to `getfield`

### LINKED ISSUE #46: `@functor` is confused about properties vs fields
It was noticed in https://github.com/FluxML/Flux.jl/issues/2107 that Functors.jl + ProtoStruct.jl doesn't work, as Functors uses `fieldnames` + `getproperty`, which is overloaded by ProtoStruct.jl. 

https://github.com/FluxML/Functors.jl/blob/v0.3.0/src/functor.jl#L11-L16

This is a bug, Functors should use `getfield` to be consistent. (Possibly the code was written before `getproperty` existed?)

mcabbott:
Closes #46

Is this going to have any surprising consequences?

Note that `@functor` cannot go the other way, and use `getproperty`/ `propertynames` for everything, as that needs an instance not the type. 

(Could add tests but they'd be kind-of contrived, I mean all existing tests check that this runs, new ones would only check against someone editing these lines back to `getproperty`.)

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -1,19 +1,19 @@
 
-functor(T, x) = (), _ -> x
+functor(T, x) = (), Returns(x)
 functor(x) = functor(typeof(x), x)
 
 functor(::Type{<:Tuple}, x) = x, identity
-functor(::Type{<:NamedTuple{L}}, x) where L = NamedTuple{L}(map(s -> getproperty(x, s), L)), identity
+functor(::Type{<:NamedTuple{L}}, x) where L = NamedTuple{L}(map(s -> getfield(x, s), L)), identity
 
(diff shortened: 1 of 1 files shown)


mcabbott:
Could do milestones. Flux seems messy enough that they help, perhaps this is simple enough not to?

mcabbott:
This breaks the following use, which relies on Tangent overloading `getproperty`:
```julia
julia> using ChainRulesCore

julia> x = (a = 1,);

julia> gx = ProjectTo(x)((a=0.1,))
Tangent{NamedTuple{(:a,), Tuple{Int64}}}(a = 0.1,)

julia> gx.a
0.1

julia> functor(typeof(x), gx)
ERROR: type Tangent has no field a
```
I think that means we should revert it. It will break Optimisers + non-Zygote completely.

PR #51, opened 2022-11-25
functor by default

### LINKED ISSUE #49: Does `functor` have the  right semantics for Flux?
Due to the default fallback 
```julia
functor(T, x) = (), _ -> x
```
in [Functors.jl](https://github.com/FluxML/Functors.jl/blob/ccdc06355c7f7c1a96bc0e9e69110936c979cc0f/src/functor.jl#L1), every custom type is considered a leaf (i.e. it has no children) and we have to sprinkle `@functor MyType` everywhere in Flux and in user code. 
We could remove all this boilerplate by having by default what `@functor MyType` currently does. Then 99% of people could live their life completely unaware of `@functor/functor` (historically poorly documented and poorly understood) and only use the much clearer `trainable(x::MyType)` in case they need to customize the parameter collection.

Besides the transition, which I think could be made rather smooth, does anyone see any counterindication in changing the default?

CarloLucibello:
Makes everything a functor by default, avoiding the need to sprinkle
```julia
@functor T
```
everywhere in Flux's layers and similar use cases.
The types already decorated with `@functor T` or `@functor T (a, b)` won't be affected by the change.

The amount of breakage and unintended consequence this PR could produce is something I cannot estimate at the moment.

Fix #49 

**TODO**:
- [x] docs
- [x] run Optimisers.jl tests on this PR
- [x] run Flux.jl tests on this PR


src/functor.jl
```diff
@@ -9,14 +17,20 @@ functor(::Type{<:AbstractArray}, x) = x, identity
...
-  escargs = map(fieldnames(T)) do f
-    f in fs ? :(y[$(yᵢ += 1)]) : :(x.$f)
-  end
-  escfs = [:($f=x.$f) for f in fs]
-
-  @eval m begin
-    $Functors.functor(::Type{<:$T}, x) = (;$(escfs...)), y -> $T($(escargs...))
+  if isempty(fs)
+    @eval m begin
+      $Functors.functor(::Type{<:$T}, x) = (), _ -> x
```
mcabbott, line 22: This explicitly reverts #36

src/functor.jl
```diff
@@ -1,5 +1,13 @@
+function functor(T, x)
+  names = fieldnames(T)
+  if isempty(names)
+    return (), _ -> x
+  end
+  S = T.name.wrapper # remove parameters from parametric types
+  vals = ntuple(i -> getfield(x, names[i]), length(names))
+  return NamedTuple{names}(vals), y -> S(y...)
+end
```
mcabbott, line 9: I think this will be slow. In https://github.com/FluxML/Flux.jl/pull/1932 it needed a generated function to be as quick as before:

https://github.com/FluxML/Flux.jl/pull/1932/files#diff-34ea0dd342eeb0864015f5ab93bd9db67f6f46ec46d4bf8dbebe9b8c1a433b04R129-R141

mcabbott (commented):
The fact that there's still a method for AbstractArray{<:Number} means that it doesn't recurse into the reshape here, which is good I think:
```julia
julia> pr(x) = (@show typeof(x); x);

julia> fmap(pr, rand(3)');
typeof(x) = Vector{Float64}

julia> fmap(pr, reshape(rand(Int8, 4)',2,2))
typeof(x) = Base.ReshapedArray{Int8, 2, Adjoint{Int8, Vector{Int8}}, Tuple{}}
2×2 reshape(adjoint(::Vector{Int8}), 2, 2) with eltype Int8:
   53  -63
 -125  -58
```
The default functor doesn't seem able to reconstruct closures like:
```julia
julia> D = let W = rand(2,2), b = zeros(2)
             x -> tanh.(W*x .+ b)
           end
#11 (generic function with 1 method)

julia> fmap(pr, D)
typeof(x) = Matrix{Float64}
typeof(x) = Vector{Float64}
ERROR: MethodError: no method matching var"#11#12"(::Matrix{Float64}, ::Vector{Float64})
Stacktrace:
 [1] (::Functors.var"#3#6"{UnionAll})(y::NamedTuple{(:W, :b), Tuple{Matrix{Float64}, Vector{Float64}}})
   @ Functors ~/.julia/packages/Functors/1AaAn/src/functor.jl:8
 [2] (::Functors.DefaultWalk)(::Function, ::Function)
   @ Functors ~/.julia/packages/Functors/1AaAn/src/walks.jl:56
...

julia> fieldnames(var"#11#12")
(:W, :b)

julia> methods(var"#11#12")
# 0 methods for type constructor
```
In global scope, example from https://fluxml.ai/Flux.jl/stable/models/basics/#Building-Simple-Models just does nothing instead, unsurprisingly:
```julia
julia> W = rand(2, 5);
julia> b = rand(2);
julia> predict(x) = W*x .+ b;

julia> fmap(pr, predict)
typeof(x) = typeof(predict)
predict (generic function with 1 method)

julia> fieldnames(typeof(predict))
()
```

src/functor.jl
CarloLucibello: Let's discuss this in #52

mcabbott:
I think Kyle said he had a branch doing something like this, using ConstructionBase. That appears to be able to reconstruct closures which is neat:
```julia
julia> adder = let y = ones(1)
                 x -> x .+ y
               end
#38 (generic function with 1 method)

julia> adder.y
1-element Vector{Float64}:
 1.0

julia> adder(2)
1-element Vector{Float64}:
 3.0

julia> using ConstructionBase

julia> newadder = constructorof(typeof(adder))([4 5 6])
#38 (generic function with 1 method)

julia> newadder(2)
1×3 Matrix{Int64}:
 6  7  8
```
It is happy to re-build things like `reshape(rand(Int8, 4)',2,2)`, not sure how that will interact with `fmap(f, x, dx)` for gradients etc.

ToucheSir:
I learned recently that it's possible to de- and reconstruct closure types. See https://github.com/JuliaGPU/Adapt.jl/pull/58 for an implementation of this. Even if we can't functor all user-defined types by default, maybe this is something we could in a backwards-compatible way? I could see it as a pilot project of sorts too.

src/functor.jl
CarloLucibello: let's keep this in mind for a future optimization-oriented PR

CarloLucibello:
With Flux v0.15 in preparation I think this is the right time to merge this.

src/functor.jl
CarloLucibello: @mcabbott I'm not familiar with generated function. Can you show me the snippet we should use here? Otherwise it will be done in a future PR.

CarloLucibello:
I marked as **leaves**
- `Number`
- `AbstractArray{<:Number}`
- `AbstractString`
Other suggestions?

mcabbott:
Would be good for docs to note that not every AbstractArray of Numbers is a leaf, e.g. Transpose

CarloLucibello:
If I can get an approval I would go on and merge this

src/functor.jl
mcabbott: My link above is a few lines off. I don't remember all the details, but I think my comments are saying that it works around `Base.typename(T).wrapper` taking 2μs every time. However, this PR doesn't do that, it calls `S = constructorof(T)`. Maybe that removes the need for `@generated`. Might be worth timing these two?

mcabbott: Trying to time this today...
```julia
julia> struct Mine1{T,S}
       x::T
       y::S
       end

julia> x = Mine1([1,2,3], [4,5,6]')
Mine1{Vector{Int64}, Adjoint{Int64, Vector{Int64}}}([1, 2, 3], adjoint([4, 5, 6]))

julia> @btime Base.typename(typeof($x)).wrapper
  1.458 ns (0 allocations: 0 bytes)
Mine1

julia> @btime ConstructionBase.constructorof(typeof($x))
  1.458 ns (0 allocations: 0 bytes)
Mine1

julia> VERSION
v"1.12.0-DEV.1375"
```
and on Julia 1.6 -- slower but far from 2μs, even if I have a faster computer than I did then...
```julia
julia> @btime Base.typename(typeof($x)).wrapper
  14.194 ns (0 allocations: 0 bytes)
Mine1

julia> @btime ConstructionBase.constructorof(typeof($x))
  0.001 ns (0 allocations: 0 bytes)
Mine1

julia> VERSION
v"1.6.0"
```

mcabbott:
Can this update the readme too? I think that (besides deleting `@functor Foo`) it wants a big notice that Functors@0.5 is sort-of opt-out now, defaults to calling ConstructionBase.

mcabbott (approved):
Let's do it.

CarloLucibello:
let's unleash havoc on the world!
ahah, hopefully not

PR #53, opened 2022-11-25
functorize Dict

### LINKED ISSUE #45: Support dictionary
### Describe the potential feature

Currently, `xs, re = Functors.functor(Dict("a"=>1.0))` returns empty for `xs`

### Motivation

Support complex models using Dictionary. Related issue: https://github.com/FluxML/Optimisers.jl/issues/114

### Possible Implementation

_No response_

### LINKED ISSUE #52: empty namedtuple is not a leaf
```julia
julia> Functors.isleaf((;))
false
```
I guess we should consider this a bug.

CarloLucibello:
With this PR we treat `Dict` similarly to namedtuples: `fmap` involves only the values.
`Base.map` shied away from taking a stance on how to behave with dictionaries and throws an error, so I had to define an internal `_map` method.

Fix #45

### DIFF
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -111,8 +111,8 @@ A structure and type preserving `map`.
 
 By default it transforms every leaf node (identified by `exclude`, default [`isleaf`](@ref))
 by applying `f`, and otherwise traverses `x` recursively using [`functor`](@ref).
-Optionally, it may also be associated with obejcts `ys` with the same tree structure.
-In that case, `f` is applied to the corresponding leaff nodes in `x` and `ys`.
+Optionally, it may also be associated with objects `ys` with the same tree structure.
+In that case, `f` is applied to the corresponding leaf nodes in `x` and `ys`.
 
 # Examples
 ```jldoctest
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -111,6 +111,8 @@ A structure and type preserving `map`.
 
 By default it transforms every leaf node (identified by `exclude`, default [`isleaf`](@ref))
 by applying `f`, and otherwise traverses `x` recursively using [`functor`](@ref).
+Optionally, it may also be associated with obejcts `ys` with the same tree structure.
+In that case, `f` is applied to the corresponding leaff nodes in `x` and `ys`.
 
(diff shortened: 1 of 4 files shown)


DhairyaLGandhi:
We can already accumulate over dictionaries. Is there a reason why we would
not follow the Base pattern?

CarloLucibello:
> We can already accumulate over dictionaries

that is not the point

>  Is there a reason why we would not follow the Base pattern?

I don't know what that means.

The use for this in the flux ecosystem is to allow models to store parameters inside dictionaries

src/walks.jl
```diff
@@ -1,3 +1,7 @@
+_map(f, x...) = map(f, x...)
+_map(f, x::Dict) = Dict(k => f(v) for (k, v) in x)
```
ToucheSir, line 2: What happens if we get `_map(f, x::Dict...)`? Currently it would result in a MethodError from inside `map`, so ideally we'd at least make a more user friendly message.

CarloLucibello: I extended the method to support multiple dictionaries

src/Functors.jl
```diff
@@ -143,6 +145,13 @@ julia> fmap(println, (i = twice, ii = 34, iii = [5, 6], iv = (twice, 34), v = 34
 34
 34.0
 (i = nothing, ii = nothing, iii = nothing, iv = (nothing, nothing), v = nothing)
+
+julia> d1 = Dict("x" => [1,2], "y" => 3);
+
+julia> d2 = Dict("x" => [4,5], "y" => 6);
+
+julia> fmap(+, d1, d2) == Dict("x" => [5, 7], "y" => 9)
```
ToucheSir, line 153: Since `_map(f, x::Dict, ys...)` mandates all `ys` have a superset of the keys `x` does and ignores any extra ones, modifying this example to reflect that would be good. Alternatively, we could add a `!!!` warning box in the docs about this behaviour with a separate example.

CarloLucibello: I modified the example to show the behavior. Also added an extra test.

ToucheSir (approved):

PR #54, opened 2022-11-25
isleaf is isempty

### LINKED ISSUE #52: empty namedtuple is not a leaf

CarloLucibello:
implement the suggestion in https://github.com/FluxML/Functors.jl/issues/52#issuecomment-1327734936

Fix #52

### DIFF
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -71,7 +71,7 @@ var"@functor"
     Functors.isleaf(x)
 
 Return true if `x` has no [`children`](@ref) according to [`functor`](@ref).
-Corresponding to `isempty(children(x)) == true`.
+Corresponds to `isempty(children(x)) == true`.
 
 # Examples
 ```jldoctest
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -84,7 +84,7 @@ julia> Functors.isleaf(["five", [6, 7]])
 false
 
 julia> Functors.isleaf([])
-false
+true
 
 julia> Functors.isleaf((8, 9))
 false
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -71,6 +71,7 @@ var"@functor"
     Functors.isleaf(x)
 
 Return true if `x` has no [`children`](@ref) according to [`functor`](@ref).
+Corresponding to `isempty(children(x)) == true`.
 
 # Examples
 ```jldoctest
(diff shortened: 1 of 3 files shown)


test/basics.jl
```diff
@@ -149,6 +149,14 @@ end
   @test_throws Exception functor(NamedTuple{(:x, :y)}, (z=33, x=1))
 end
 
+@testset "empty children" begin
+  struct EmptyChildren; a; b; end
+  @functor EmptyChildren ()
+  x = EmptyChildren(1, 2)
+  @test children(x) === (;)
```
CarloLucibello, line 156:
```suggestion
  @test Functors.children(x) === (;)
```

test/basics.jl
```diff
@@ -149,6 +149,14 @@ end
...
+  @test isleaf(x)
```
CarloLucibello, line 157:
```suggestion
  @test Functors.isleaf(x)
```

src/Functors.jl
```diff
@@ -71,6 +71,7 @@ var"@functor"
     Functors.isleaf(x)
 
 Return true if `x` has no [`children`](@ref) according to [`functor`](@ref).
+Corresponding to `isempty(children(x)) == true`.
```
CarloLucibello, line 74:
```suggestion
Corresponds to `isempty(children(x)) == true`.
```

CarloLucibello:
I don't see why we shouldn't do this. I think that `[]`, `()`, `(;)` can be safely considered as leafs

ToucheSir:
It would be pretty breaking, because we use `exclude = isleaf` to decide what to pass to map function callbacks. Code that never expected to encounter an empty (non-numeric) array or dict before is now going to have to deal with them. At a higher level, making two possible paths for particular types based on their contents over just types blows up any chance of us making traversals type stable and introduces additional overhead for users (who will now have to detect and guard against random empty arrays/dicts which they're not interested in. Distinguishing those from parameters would not be fun!)

As I mentioned in https://github.com/FluxML/Functors.jl/issues/52#issuecomment-1327980603, the safer path seems to be limiting and not broadening the scope of `isleaf` so that we have fewer priveleged types.

CarloLucibello:
right, I clearly didn't think it through.

PR #55, opened 2022-11-29
add `@leaf` macro

CarloLucibello:
and use the alias `NoChildren` for `Tuple{}`.
No breaking change and no great utility at the moment either, but `@leaf` will be useful when/if the functor structure will be opt-out (#51).

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -1,5 +1,5 @@
 
-functor(T, x) = (), _ -> x
+@leaf Any
 functor(x) = functor(typeof(x), x)
 
 functor(::Type{<:Tuple}, x) = x, identity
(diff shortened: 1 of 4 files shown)


src/functor.jl
```diff
@@ -1,13 +1,31 @@
+function functor end
 
-functor(T, x) = (), _ -> x
+const NoChildren = Tuple{}
+
+function makeleaf(m::Module, T)
+  @eval m begin
```
ToucheSir, line 6: If I'm not mistaken, we should be able to do this without the `@eval` by emitting the function definition from the macro directly.

CarloLucibello: I need more indirection for that, my macro skills are terrible, I'm almost blindly adapting the code from `@functor`

ToucheSir: Mine are pretty atrocious as well, but after playing around locally this might work:
```julia
macro leaf(T)
  :($Functors.functor(::Type{<:$(esc(T))}, x) = $Functors.NoChildren(), _ -> x
end
```

CarloLucibello:
```
macro leaf(T)
  :($Functors.functor(::Type{<:$(esc(T))}, x) = $Functors.NoChildren(), _ -> x)
end
```
errors with 
```julia
julia> using Functors
[ Info: Precompiling Functors [d9f16b24-f501-4c13-a1f2-28368ffc5196]
ERROR: LoadError: syntax: invalid named tuple field name "Functors.functor(::Type{<:(escape Any)}, x)" around /Users/carlo/.julia/dev/Functors/src/functor.jl:20
```

CarloLucibello: works

CarloLucibello:
merge?

ToucheSir (approved):

Issue #59, opened 2023-01-15
Recursive inference failure for multiple functors


gaurav-arya:
I've been using `Functors` in a case where I want to make sure it's really fast in the special case of small tuples. For a single functor, this is fine, but for functions of multiple functors...
```
julia> using Functors
julia> @btime Functors.fmap(+, (3,))
  0.859 ns (0 allocations: 0 bytes)
(3,)

julia> @btime Functors.fmap(+, (3,), (3,))
  404.070 ns (9 allocations: 464 bytes)
(6,)
```
JET.jl states the following for the second call
```
═════ 7 possible errors found ═════
┌ @ /home/gaurav/.julia/dev/Functors/src/maps.jl:3 Functors.:(var"#fmap#134")(tuple(Functors.isleaf, Functors.DefaultWalk(), Functors.IdDict(), Functors.NoKeyword(), #self#, f, x), ys...)
│┌ @ /home/gaurav/.julia/dev/Functors/src/maps.jl:11 fmap(tuple(_walk, f, x), ys...)
││┌ @ /home/gaurav/.julia/dev/Functors/src/maps.jl:1 walk(tuple(#132, x), ys...)
│││┌ @ /home/gaurav/.julia/dev/Functors/src/walks.jl:132 ret = walk.walk(tuple(recurse, x), ys...)
││││┌ @ /home/gaurav/.julia/dev/Functors/src/walks.jl:92 walk.walk(tuple(recurse, x), ys...)
│││││┌ @ /home/gaurav/.julia/dev/Functors/src/walks.jl:62 Functors._map(tuple(recurse, func), yfuncs...)
││││││┌ @ /home/gaurav/.julia/dev/Functors/src/walks.jl:1 Functors.map(tuple(f), x...)
│││││││┌ @ tuple.jl:298 f(t[1], s[1])
││││││││┌ @ /home/gaurav/.julia/dev/Functors/src/maps.jl:1 fmap(tuple(getfield(#self#, :walk), getfield(#self#, :f)), xs...)
│││││││││┌ @ /home/gaurav/.julia/dev/Functors/src/maps.jl:1 Functors.fmap(::Functors.CachedWalk{ExcludeWalk{DefaultWalk, typeof(+), typeof(Functors.isleaf)}, Functors.NoKeyword}, ::typeof(+), ::Int64, ::Int64)
[...]

PR #61, opened 2023-01-16
Fix recursion performance

gaurav-arya:
This PR simplifies the recursive structure of walking a functor. It uses a trick from https://github.com/JuliaLang/julia/issues/47760 to resolve https://github.com/FluxML/Functors.jl/issues/59. 

The functionality of `fmap` after this PR is identical to its previous one. Thus, this PR lays bare the fact that the argument `f` is ultimately useless for the version of `fmap` that takes a user-specified walk. Note that this was public API, however, so I kept it in the signature and added an explanatory note.

Obviously the `fmap` API is highly imperfect, but I think this PR is a strict improvement, that simply lays bare some of the issues. (Ultimately, I think that the use of `fmap` with a user-specified arbitrary walk should probably be deprecated: walks are just way too flexible to allow arbitrary walks to be paired with `fmap`. The semantics of `fmap` seem to match best with the specific use case of `ExcludeWalk`: `ExcludeWalk` is what actually uses `f`, after all. The general function allowing for arbitrary walks should probably be called `recursivewalk` or something. But that would be very breaking.)

### DIFF
--- a/src/maps.jl
+++ b/src/maps.jl
@@ -1,4 +1,12 @@
-fmap(walk::AbstractWalk, f, x, ys...) = walk((xs...) -> fmap(walk, f, xs...), x, ys...)
+function fmap(walk::AbstractWalk, f, x, ys...)
+  #=
+  The below construct avoids a performance penalty
+  for recursive constructs in an anonymous function.
+  See Julia issue #47760 and Functors.jl issue #59.
+  =#
+  recurse(xs...) = walk(var"#self#", xs...)
+  walk(recurse, x, ys...)
+end
 
 function fmap(f, x, ys...; exclude = isleaf,
                            walk = DefaultWalk(),
--- a/src/maps.jl
+++ b/src/maps.jl
@@ -1,17 +1,8 @@
-#=
-Note that the argument f is not actually used in 
-the below version of fmap, which takes a user-specified
-walk. Since this function is public API, the f argument
-has been left in the signature until a (likely breaking)
-decision can be made in the future about how to clean up
-this API (issue #62).
-=#
+# Note that the argument f is not actually used in this method.
+# See issue #62 for a discussion on how best to remove it.
 function fmap(walk::AbstractWalk, f, x, ys...)
-  #=
-  The below construct avoids a performance penalty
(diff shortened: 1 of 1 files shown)


darsnack (approved):
Are there packages outside Flux using the walk-based API? Better to make a breaking change soon before more people adopt the new (bad) API.

src/maps.jl
```diff
@@ -1,4 +1,20 @@
...
+this API (issue #62).
+=#
+function fmap(walk::AbstractWalk, f, x, ys...)
+  #=
+  The below construct avoids a performance penalty
+  for recursive constructs in an anonymous function.
+  See Julia issue #47760 and Functors.jl issue #59.
+  =#
+  recurse(xs...) = walk(var"#self#", xs...)
+  walk(recurse, x, ys...)
```
ToucheSir, line 16:
```suggestion
  # This avoids a performance penalty for recursive constructs in an anonymous function.
  # See Julia issue #47760 and Functors.jl issue #59.
  recurse(xs...) = walk(var"#self#", xs...)
  walk(recurse, x, ys...)
```

src/maps.jl
```diff
@@ -1,4 +1,20 @@
-fmap(walk::AbstractWalk, f, x, ys...) = walk((xs...) -> fmap(walk, f, xs...), x, ys...)
+#=
+Note that the argument f is not actually used in 
+the below version of fmap, which takes a user-specified
+walk. Since this function is public API, the f argument
+has been left in the signature until a (likely breaking)
+decision can be made in the future about how to clean up
+this API (issue #62).
+=#
```
ToucheSir, line 8:
```suggestion
# Note that the argument f is not actually used in this method.
# See issue #62 for a discussion on how best to remove it.
```

ToucheSir (approved):
Thanks! Just a couple of formatting nits from me.

gaurav-arya:
I don't believe this has got a release?

ToucheSir:
I don't remember why. Might've held off because of the recent TagBot issues that prevented tags and release notes from being published. See https://github.com/JuliaRegistries/General/pull/77862

PR #63, opened 2023-01-19
Adjustments to walk recursion

### LINKED ISSUE #62: fmap API that accepts an arbitrary walk does not use function f

gaurav-arya:
Adjusts the `walk::AbstractWalk` interface to naturally support recursion through `walk(x, ys...)` and deprecates an old `fmap` signature; resolves #62.

One subtle point: I wanted to leave the `recurse` first-arg exactly as is which would have made the diff even smaller. However, for the dispatch to work, we needed to restrict the first arg `recurse` to be an `AbstractWalk` when implementing walk subtypes, in which case it made more sense to call it `outer_walk`, or something of that nature. [This is a breaking change for people who have defined custom walks, since custom subtypes of walk will need to make sure to have this dispatch. This leads to IMO the cleanest API, but I could use something more clunky than `walk(x, ys...)` for the recursive walk and avoid the breaking change if desired. The main point of this PR is to show that something like the  `recurse` closure is not necessary]

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Functors"
 uuid = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
-version = "0.4.2"
+version = "0.5.0"
 
 [deps]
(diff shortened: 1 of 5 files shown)


gaurav-arya:
Changing how custom walks are defined would be really breaking for `Optimisers`, it seems. Closing and trying again.

PR #64, opened 2023-01-19
 Deprecate fmap(walk, f, x, ys...) for simpler runwalk function

### LINKED ISSUE #62: fmap API that accepts an arbitrary walk does not use function f

gaurav-arya:
The new structure avoids the need for a recursive closure. Resolves #62 

Other than the deprecation, this should not be breaking according to the public API (and tests pass unchanged), although it should definitely be checked in practice.


gaurav-arya:
Crap, this is probably breaking too due to the change in how `recurse` is used.

PR #65, opened 2023-01-19
Deprecate fmap(walk, f, x, ys...) for execute(walk, x, ys...)

### LINKED ISSUE #62: fmap API that accepts an arbitrary walk does not use function f

gaurav-arya:
Deprecates `fmap(walk, f, x, ys...)` for `runwalk(walk, x, ys...)`, with no higher ambitions. Resolves #62 

Edit: `runwalk` now renamed to `execute`

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -12,6 +12,7 @@ Functors.isleaf
 
 ```@docs
 Functors.AbstractWalk
+Functors.runwalk
 Functors.DefaultWalk
 Functors.StructuralWalk
 Functors.ExcludeWalk
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -12,7 +12,7 @@ Functors.isleaf
 
 ```@docs
 Functors.AbstractWalk
-Functors.runwalk
+Functors.execute
 Functors.DefaultWalk
 Functors.StructuralWalk
 Functors.ExcludeWalk
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -1,6 +1,6 @@
 module Functors
 
-export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect
+export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, runwalk
 
 include("functor.jl")
 include("walks.jl")
@@ -105,7 +105,6 @@ children
 
 """
     fmap(f, x, ys...; exclude = Functors.isleaf, walk = Functors.DefaultWalk()[, prune])
-    fmap(walk, f, x, ys...)
 
 A structure and type preserving `map`.
 
@@ -190,10 +189,10 @@ use [`fmapstructure`](@ref).
 
 For advanced customization of the traversal behaviour,
 pass a custom `walk` function that subtypes [`Functors.AbstractWalk`](ref).
-The form `fmap(walk, f, x, ys...)` can be called for custom walks.
-The simpler form `fmap(f, x, ys...; walk = mywalk)` will wrap `mywalk` in
+The call `fmap(f, x, ys...; walk = mywalk)` will wrap `mywalk` in
 [`ExcludeWalk`](@ref) then [`CachedWalk`](@ref).
-
+Here, [`ExcludeWalk`](@ref) is responsible for applying `f` at excluded nodes.
+For a barebones interface that gives the user full control of the walk, see [`runwalk`](@ref).
 ```jldoctest withfoo
 julia> struct MyWalk <: Functors.AbstractWalk end
 
@@ -202,9 +201,6 @@ julia> (::MyWalk)(recurse, x) = x isa Bar ? "hello" :
 
 julia> fmap(x -> 10x, m; walk = MyWalk())
 Foo("hello", (40, 50, "hello"))
-
-julia> fmap(MyWalk(), x -> 10x, m)
-Foo("hello", (4, 5, "hello"))
 ```
 
 The behaviour when the same node appears twice can be altered by giving a value
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -1,6 +1,6 @@
 module Functors
 
-export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, runwalk
+export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, execute
 
 include("functor.jl")
 include("walks.jl")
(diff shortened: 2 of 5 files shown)


gaurav-arya:
Closing per discussion in #62

gaurav-arya:
Well, reopening in case the little bit of work I did with the deprecation becomes something we want to use in the future, but this PR isn't ready for merge.

darsnack:
I think we should go ahead and do this, but use `traverse` and note in the docstring that this is post-order traversal.

Just getting rid of the confusing `fmap(walk, f, ...)` method will be an improvement even if the naming isn't perfect. I imagine a proper renaming that mirrors prior literature on "reduce," "traverse," "fold," "map," etc. will be bigger, breaking overhaul anyways.

gaurav-arya:
Sounds good! Happy to do a renaming. I'm musing about whether `traverse` is the appropriate name, though. I don't speak Haskell, but from https://stackoverflow.com/questions/7460809/can-someone-explain-the-traverse-function-in-haskell it looks like it's an `fmap` that guarantees a traversal ordering and allows effectful functions. In particular, it seems like it should definitely have the same signature as `fmap` (and the stack overflow answer notes that the distinction between `fmap` and `traverse` is blurred in an impure language.)

So maybe `traverse` isn't the best name for this function? To me, it feels like this a lower-level primitive which is very much to do with the specific concept of a walk in this library, So maybe something like `dowalk` to avoid the oxymoron? Very happy to be contradicted and for other suggestions in the name.

darsnack:
How about `descend(walk, x, ys...)` (or `execute` or `enter`)?

CarloLucibello:
@gaurav-arya let's decide a name and merge this, if you have some time

docs/src/api.md
```diff
@@ -12,6 +12,7 @@ Functors.isleaf
 
 ```@docs
 Functors.AbstractWalk
+Functors.runwalk
```
gaurav-arya, line 15:
```suggestion
Functors.execute
```

src/Functors.jl
```diff
@@ -1,6 +1,6 @@
 module Functors
 
-export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect
+export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, runwalk
```
gaurav-arya, line 3:
```suggestion
export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, execute
```

src/Functors.jl
```diff
@@ -190,10 +189,10 @@ use [`fmapstructure`](@ref).
 
 For advanced customization of the traversal behaviour,
 pass a custom `walk` function that subtypes [`Functors.AbstractWalk`](ref).
-The form `fmap(walk, f, x, ys...)` can be called for custom walks.
-The simpler form `fmap(f, x, ys...; walk = mywalk)` will wrap `mywalk` in
+The call `fmap(f, x, ys...; walk = mywalk)` will wrap `mywalk` in
 [`ExcludeWalk`](@ref) then [`CachedWalk`](@ref).
-
+Here, [`ExcludeWalk`](@ref) is responsible for applying `f` at excluded nodes.
+For a barebones interface that gives the user full control of the walk, see [`runwalk`](@ref).
```
gaurav-arya, line 195:
```suggestion
For a low-level interface for executing a user-constructed walk, see [`execute`](@ref).
```

src/maps.jl
```diff
@@ -1,11 +1,4 @@
-# Note that the argument f is not actually used in this method.
-# See issue #62 for a discussion on how best to remove it.
-function fmap(walk::AbstractWalk, f, x, ys...)
-  # This avoids a performance penalty for recursive constructs in an anonymous function.
-  # See Julia issue #47760 and Functors.jl issue #59.
-  recurse(xs...) = walk(var"#self#", xs...)
-  walk(recurse, x, ys...)
-end
+Base.@deprecate fmap(walk::AbstractWalk, f, x, ys...) runwalk(walk, x, ys...)
```
gaurav-arya, line 1:
```suggestion
Base.@deprecate fmap(walk::AbstractWalk, f, x, ys...) execute(walk, x, ys...)
```

src/maps.jl
```diff
@@ -15,10 +8,10 @@ function fmap(f, x, ys...; exclude = isleaf,
   if !isnothing(cache)
     _walk = CachedWalk(_walk, prune, cache)
   end
-  fmap(_walk, f, x, ys...)
+  runwalk(_walk, x, ys...)
```
gaurav-arya, line 11:
```suggestion
  execute(_walk, x, ys...)
```

src/maps.jl
```diff
@@ -15,10 +8,10 @@ function fmap(f, x, ys...; exclude = isleaf,
...
 end
 
 fmapstructure(f, x; kwargs...) = fmap(f, x; walk = StructuralWalk(), kwargs...)
 
 fcollect(x; exclude = v -> false) =
-  fmap(ExcludeWalk(CollectWalk(), _ -> nothing, exclude), _ -> nothing, x)
+  runwalk(ExcludeWalk(CollectWalk(), _ -> nothing, exclude), x)
```
gaurav-arya, line 17:
```suggestion
  execute(ExcludeWalk(CollectWalk(), _ -> nothing, exclude), x)
```

src/walks.jl
```diff
@@ -24,6 +24,19 @@ The choice of which nodes to recurse and in what order is custom to the walk.
 """
 abstract type AbstractWalk end
 
+"""
+    runwalk(walk, x, ys...)
```
gaurav-arya, line 28:
```suggestion
    execute(walk, x, ys...)
```

src/walks.jl
```diff
@@ -24,6 +24,19 @@ The choice of which nodes to recurse and in what order is custom to the walk.
...
+Run a `walk` that recurses on itself, starting at a node `x` in a Functors tree,
```
gaurav-arya, line 29:
```suggestion
Execute a `walk` that recursively calls itself, starting at a node `x` in a Functors tree,
```

src/walks.jl
```diff
@@ -24,6 +24,19 @@ The choice of which nodes to recurse and in what order is custom to the walk.
...
+as well as optional associated nodes `ys...` in other Functors trees.
+Any custom `walk` function that subtypes [`Functors.AbstractWalk`](@ref) is permitted.
+"""
+function runwalk(walk::AbstractWalk, x, ys...)
```
gaurav-arya, line 33:
```suggestion
function execute(walk::AbstractWalk, x, ys...)
```

gaurav-arya (commented):
Sure -- have gone with `execute`

src/walks.jl
```diff
@@ -24,6 +24,19 @@ The choice of which nodes to recurse and in what order is custom to the walk.
 """
 abstract type AbstractWalk end
 
+"""
+    execute(walk, x, ys...)
+Execute a `walk` that recursively calls itself, starting at a node `x` in a Functors tree,
```
darsnack, line 29:
```suggestion

Execute a `walk` that recursively calls itself, starting at a node `x` in a Functors tree,
```

darsnack (approved):

darsnack:
Thank you @gaurav-arya for all the hard work and thoughts here!

Issue #68, opened 2023-05-07
`functor` reconstructs unexpectedly due to the lack of checking NameTuple's keys.


skyleaworlder:
```jl
julia> using Functors

julia> struct C
       a
       b
       c
       end

julia> @functor C (c, a)
Tuple{Expr, Expr, Expr}(:(y[1]), :(x.b), :(y[2]))
Vector{Expr}Expr[:(c = x.c), :(a = x.a)]

julia> c = C(1,2,3)
C(1, 2, 3)

julia> Functors.functor(c)
((c = 3, a = 1), var"#5#6"{C}(C(1, 2, 3)))

julia> cc, re = Functors.functor(c)
((c = 3, a = 1), var"#5#6"{C}(C(1, 2, 3)))

julia> re(map(float, cc))
C(3.0, 2, 1.0)
```

I think it's because `makefunctor` only use "y[$(y, += 1)]", ignoring the sequence of fields and the information of NameTuple's keys:

```jl
function makefunctor(m::Module, T, fs = fieldnames(T))
  yᵢ = 0
  escargs = map(fieldnames(T)) do f
    f in fs ? :(y[$(yᵢ += 1)]) : :(x.$f)
  end
  escfs = [:($f=x.$f) for f in fs]

  @eval m begin
    $Functors.functor(::Type{<:$T}, x) = (;$(escfs...)), y -> $T($(escargs...))
  end
end
```

Or this is actually nothing, because it is required to pass in order, but docs is needed.

PR #70, opened 2023-07-08
Preserve `@functor` property order when reconstructing

ToucheSir:
Fixes https://github.com/FluxML/Functors.jl/issues/68.

I originally hoped to have just the NamedTuple overload, but it turns out we're still documenting reconstructing from a plain tuple: https://github.com/FluxML/Functors.jl/blob/v0.4.4/src/Functors.jl#L49-L62.

### PR Checklist

- [x] Tests are added
~~- [ ] Documentation, if applicable~~

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -22,14 +22,21 @@ functor(::Type{<:AbstractArray}, x) = x, identity
 @leaf AbstractArray{<:Number}
 
 function makefunctor(m::Module, T, fs = fieldnames(T))
-  yᵢ = 0
+  fidx = Ref(0)
   escargs = map(fieldnames(T)) do f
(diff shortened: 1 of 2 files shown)


skyleaworlder (approved):

CarloLucibello (approved):

skyleaworlder:
As for plain tuple, it can be done using `escfs` like this (although `assign_expr.args[1]` might be a bit tricky):

```julia
      reconstruct(y) = begin 
          pos_val = Dict([(pos => val) for (pos, val) in enumerate(y)])
          pos_name = Dict([(pos => assign_expr.args[1]) for (pos, assign_expr) in enumerate($(escfs))])
          reconstruct(NamedTuple([(name, pos_val[pos]) for (pos, name) in pos_name]))
      end
```

PR #71, opened 2023-08-26
`@functor Returns`

mcabbott:
Recently wanted this, and it seems natural given Fix1 etc.

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -10,6 +10,8 @@
 @functor Base.Fix2
 @functor Base.Broadcast.BroadcastFunction
 
+@functor Base.Returns
+
 ###
 ### Array wrappers
 ###
--- a/src/base.jl
+++ b/src/base.jl
@@ -10,7 +10,9 @@
 @functor Base.Fix2
 @functor Base.Broadcast.BroadcastFunction
 
-@functor Base.Returns
+@static if VERSION >= v"1.7"
+  @functor Base.Returns
+end
 
 ###
 ### Array wrappers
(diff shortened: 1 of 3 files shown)


ToucheSir (commented):
Needs either a `Compat` dep or a version check to make 1.6 CI happy.

src/base.jl
```diff
@@ -10,6 +10,8 @@
 @functor Base.Fix2
 @functor Base.Broadcast.BroadcastFunction
 
+@functor Base.Returns
```
ToucheSir, line 13:
```suggestion
@static if VERSION >= v"1.7"
  @functor Base.Returns
end
```

mcabbott: Thanks! I got distracted, think this is cleaner than adding Compat.jl

test/base.jl
```diff
@@ -50,6 +50,12 @@ end
   @test fmap(x -> x + 10, bf) == Base.Broadcast.BroadcastFunction(Bar(13.3))
 end
 
+@testset "Returns" begin
+  ret = Returns([0, pi, 2pi])
+  @test Functors.functor(ret)[1] == (value = [0, pi, 2pi],)
+  @test Functors.functor(ret)[2]((value = 1:3,)) === Returns(1:3)
+end
```
ToucheSir, line 57: Forgot this, I think it needs a VERSION guard too. If that's enough to pass, merge when ready.

ToucheSir (approved):

test/base.jl
mcabbott:
```suggestion
VERSION >= v"1.7" && @testset "Returns" begin
  ret = Returns([0, pi, 2pi])
  @test Functors.functor(ret)[1] == (value = [0, pi, 2pi],)
  @test Functors.functor(ret)[2]((value = 1:3,)) === Returns(1:3)
end
```

PR #72, opened 2024-02-06
Fix anymutable errors with self-referencing types

oschulz:
I currently have trouble using `fmap` with data structures that contains `Measurements.Measurement` objects. The reason is that `Measurement` (which subtypes `Real`) contains a field `Derivatives`:

```julia
struct Derivatives{T} <: AbstractDict{Tuple{T,T,UInt64},T}
    parent::Derivatives{T}
    ...
end
```

Due to the nature of the type, this `parent` field may be undefined - quite normal for such datatypes.(`Measurements.Derivatives` is a variation of `Base.ImmutableDict`, apparently).

This causes `Functors.anymutable` to fail, due to

```julia
@generated function anymutable(x::T) where {T}
  ...
  subs =  [:(anymutable(getfield(x, $f))) for f in QuoteNode.(fieldnames(T))]
  ...
end
```

While I have this problem with `Measurement` specifically, I think it makes sense to fix it here, instead of proposing a Functors extension to Measurements. Also, as expected (see above) the same problem occurs with `Base.ImmutableDict`:

```julia
julia> fmap(identity, Base.ImmutableDict(:b => 42))
ERROR: UndefRefError: access to undefined reference
```

A simple fix would be checking with `isdefined` (if the field is not defined, it obviously doesn't point to mutable content):

```julia
@generated function anymutable(x::T) where {T}
  ...
  subs =  [:(isdefined(x, $f) && anymutable(getfield(x, $f))) for f in QuoteNode.(fieldnames(T))]
  ...
end
```

I worry that this may reduce the type stability of `anymutable` though. So in this PR I'm proposing

```julia
@generated function anymutable(x::T) where {T}
  ...
  fns = QuoteNode.(filter(n -> fieldtype(T, n) != T, fieldnames(T)))
  subs =  [:(anymutable(getfield(x, $f))) for f in fns]
  ...
end
```

In the end, mutability is a type property, I'd say:  if `T` itself is not directly mutable, and all of it's fields except references to `T` are immutable, then I think it's safe to consider `x` immutable as a whole.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Functors"
 uuid = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
-version = "0.4.6"
+version = "0.4.7"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,7 @@ version = "0.4.6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 
 [compat]
-Documenter = "0.27"
+Documenter = "1"
 julia = "1.6"
 
 [extras]
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,2 +1,5 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+
+[compat]
+Documenter = "1"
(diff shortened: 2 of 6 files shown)


oschulz:
Also fixes the broken docs generation (was not Documenter v1 compatible, but Documenter version bound was missing in "docs/Project.toml"). I updated everything to Documenter v1 now.

ToucheSir (approved):
I didn't quite understand what this was doing yesterday, but with a fresh set of eyes it makes sense now :)

oschulz:
Whoops, sorry @ToucheSir , I had overlooked the fact that the version number had already been increased, so https://github.com/JuliaRegistries/General/pull/100422 is currently stalled.

Could you call Registrator on commit 19788eaf785ff4b29e so that we can get v0.4.6 registered first?

PR #73, opened 2024-03-08
add function `fleaves`

CarloLucibello:
Adds utility function ~fflatten~ `fleaves` that flatten the leaves into an array, similarly to jax's [tree_leaves](https://jax.readthedocs.io/en/latest/_autosummary/jax.tree_util.tree_leaves.html).
Useful for models' parameters collection:
```julia
julia> using Flux

julia> m = Chain(Dense(2=>3, relu), Dense(3 => 2))
Chain(
  Dense(2 => 3, relu),                  # 9 parameters
  Dense(3 => 2),                        # 8 parameters
)                   # Total: 4 arrays, 17 parameters, 324 bytes.

julia> fleaves(m)
6-element Vector{Any}:
 Float32[-0.0102046095 -0.28821245; 0.4982538 0.23464581; -0.70785844 0.3291903]
 Float32[0.0, 0.0, 0.0]
 relu (generic function with 2 methods)
 Float32[0.57762665 -0.8608131 -0.5159058; -0.54350173 0.61605966 0.8265723]
 Float32[0.0, 0.0]
 identity (generic function with 1 method)
```

### DIFF
--- a/src/walks.jl
+++ b/src/walks.jl
@@ -234,7 +234,17 @@ julia> collect(zipped_iter)
 struct IterateWalk <: AbstractWalk end
 
 function (walk::IterateWalk)(recurse, x, ys...)
-  func, _ = functor(x)
-  yfuncs = map(y -> functor(typeof(x), y)[1], ys)
-  return Iterators.flatten(_map(recurse, func, yfuncs...))
+  x_children = children(x)
+  ys_children = map(children, ys)
+  return Iterators.flatten(_map(recurse, x_children, ys_children...))
 end
+
+struct FlattenWalk <: AbstractWalk end
+
+function (walk::FlattenWalk)(recurse, x, ys...)
+  x_children = _values(children(x))
+  ys_children = map(children, ys)
+  res = _map(recurse, x_children, ys_children...)
+  return reduce(vcat, _values(res))
+end
+
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -25,5 +25,5 @@ Functors.IterateWalk
 ```@docs
 Functors.fmapstructure
 Functors.fcollect
-Functors.fflatten
+Functors.fleaves
 ```
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -25,4 +25,5 @@ Functors.IterateWalk
 ```@docs
 Functors.fmapstructure
 Functors.fcollect
+Functors.fflatten
 ```
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -1,6 +1,6 @@
 module Functors
 
-export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, execute, fflatten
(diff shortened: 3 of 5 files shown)


src/walks.jl
```diff
@@ -234,7 +234,17 @@ julia> collect(zipped_iter)
 struct IterateWalk <: AbstractWalk end
 
 function (walk::IterateWalk)(recurse, x, ys...)
-  func, _ = functor(x)
-  yfuncs = map(y -> functor(typeof(x), y)[1], ys)
-  return Iterators.flatten(_map(recurse, func, yfuncs...))
+  x_children = children(x)
+  ys_children = map(children, ys)
+  return Iterators.flatten(_map(recurse, x_children, ys_children...))
```
CarloLucibello, line 239: no functional changes here, only improved clarity

darsnack:
Implementation looks good and I'm on board with adding something like this. Only thing is that this isn't `jtu.tree_flatten`. If it were, it should return the analog to a tree def as well. I would say `jtu.tree_flatten` is more like `Optimisers.restructure` except without concatenating the leaves.

It'd be nice to have something like a real `jtu.tree_flatten`, but for now I suggest renaming this to `fleaves` to match `jtu.tree_leaves`—which is what this actually is.

CarloLucibello:
Changed name to `fleaves`

CarloLucibello:
this is ready to go

darsnack (approved):

src/walks.jl
```diff
@@ -234,7 +234,17 @@ julia> collect(zipped_iter)
...
 end
+
+struct FlattenWalk <: AbstractWalk end
+
+function (walk::FlattenWalk)(recurse, x, ys...)
+  x_children = _values(children(x))
+  ys_children = map(children, ys)
+  res = _map(recurse, x_children, ys_children...)
+  return reduce(vcat, _values(res))
```
mcabbott, line 248: Note that this works pairwise, it does not use any of the magic fast methods of reduce, unless you have a vector of objects:
```julia
julia> @eval Functors function (walk::FlattenWalk)(recurse, x, ys...)
         x_children = _values(children(x))
         ys_children = map(children, ys)
         res = _map(recurse, x_children, ys_children...)
         @show typeof(_values(res))
         return reduce(vcat, _values(res))
       end

julia> fleaves((x=(1,2,3), y=(a=4, b=5), z=6))
typeof(_values(res)) = Tuple{Vector{Int64}, Vector{Int64}, Vector{Int64}}
typeof(_values(res)) = @NamedTuple{a::Vector{Int64}, b::Vector{Int64}}
typeof(_values(res)) = @NamedTuple{x::Vector{Int64}, y::Vector{Int64}, z::Vector{Int64}}
6-element Vector{Int64}:
 1
 2
 3
 4
 5
 6

julia> @which reduce(vcat, ([1], [2], [3]))
reduce(op, itr; kw...)
     @ Base reduce.jl:490
```

PR #74, opened 2024-03-08
extend `fmapstructure` and add `KeyPath`, `fmap_with_path`, `fmapstructure_with_path`

CarloLucibello:
Adding some utilities for working with paths along trees:
- `KeyPath`
- `fmap_with_path`
- `fmapstructure_with_path`
 
Since it came for free, I also extended `fmapstructure` to handle multiple inputs.

In Jax they have KeyPath and [tree_map_with_path](https://jax.readthedocs.io/en/latest/_autosummary/jax.tree_util.tree_map_with_path.html)

TODO
- [x] add tests for `fmap_with_path`
- [x] add to docs

### DIFF
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -332,9 +332,9 @@ for more information.
 ```jldoctest
 julia> x = ([1, 2, 3], 4, (a=5, b=Dict("A"=>6, "B"=>7), c=Dict("C"=>8, "D"=>9)));
 
-julia> fexclude(kp, x) = kp == KeyPath(3, :c) || Functors.isleaf(x);
+julia> exclude(kp, x) = kp == KeyPath(3, :c) || Functors.isleaf(x);
 
-julia> fmap_with_path((kp, x) -> x isa Dict ? nothing : x.^2, x; exclude = fexclude)
+julia> fmap_with_path((kp, x) -> x isa Dict ? nothing : x.^2, x; exclude = exclude)
 ([1, 4, 9], 16, (a = 25, b = Dict("B" => 49, "A" => 36), c = nothing))
 ```
 """
@@ -373,12 +373,12 @@ julia> @functor Bar
 
 julia> struct TypeWithNoChildren; x; y; end
 
-julia> m = (a=Bar([1,2,3]), b=TypeWithNoChildren(4, 5))
+julia> m = (a = Bar([1,2,3]), b = TypeWithNoChildren(4, 5));
 
 julia> fleaves(m)
 2-element Vector{Any}:
  [1, 2, 3]
- TypeWithNoChildren(:a, :b)
+ TypeWithNoChildren(4, 5)
 ```
 """
 fleaves
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -1,8 +1,10 @@
 module Functors
 
-export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, execute, fleaves
+export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, execute, fleaves, 
+       KeyPath, fmap_with_path, fmapstructure_with_path
 
 include("functor.jl")
+include("keypath.jl")
 include("walks.jl")
 include("maps.jl")
 include("base.jl")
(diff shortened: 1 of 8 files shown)


darsnack (commented):
This is great. We've talked about adding something like `KeyPath` to give nodes an "address" for things like weight sharing in Optimisers.jl (for immutable weights).


The `KeyPath` implementation look fine. But it would be nice to simplify the implementation so that we don't have two copies of every walk style.

Ultimately, the key path only matters for `ExcludeWalk`. What if we change the API so that every walk accepts a key path always? So, now the implementation for `StructuralWalk` is the body of `StructuralWalkWithKeyPath`, and we eliminate the latter. Finally, we have `ExcludeWalk` with a boolean flag indicating whether `exclude` accepts a key path or not. If `false`, then just drop passing in the key path. Alternatively, we have `ExcludeWalk` and `ExcludeWalkWithoutPath`.

It would be a breaking change, but I think it's cleaner and more maintainable in the long run.

CarloLucibello:
> But it would be nice to simplify the implementation so that we don't have two copies of every walk style.

So from the interface perspective, the suggestion would be to go from 
```julia
fmap(x -> x.^2, model, exclude = x -> isleaf(x), walk = WalkNOTTakingKeyPath())
fmap_with_path((kp, x) -> x.^2, model, exclude = (kp, x) -> isleaf(x), walk = WalkTakingKeyPath())
```
to 
```julia
fmap(x -> x.^2, model, exclude = x -> isleaf(x), walk = WalkTakingKeyPath())
fmap_with_path((kp, x) -> x.^2, model, exclude = (kp, x) -> isleaf(x), walk = WalkTakingKeyPath())
```
It seems reasonable. (maybe has an unnecessary performance hit on some exotic scenarios? I don't know).
Honestly though, I don't feel like venturing into tagging a breaking release soonish, without global considerations on what other breaking change we might put in. And I'd like to have these features in sooner rather than later. 
Also, these walks take really few lines of code, so duplication doesn't seem too problematic.
So I would say to merge this as it is, and to start listing breaking changes proposals for next breaking release.

src/Functors.jl
```diff
@@ -222,14 +226,16 @@ fmap
 
 
 """
-    fmapstructure(f, x; exclude = isleaf)
+    fmapstructure(f, x, ys...; [exclude, prune])
```
darsnack, line 229:
```suggestion
    fmapstructure(f, x, ys...; exclude = isleaf, [prune])
```

src/Functors.jl
```diff
@@ -303,4 +309,46 @@ julia> fcollect(m, exclude = v -> Functors.isleaf(v))
 """
 fcollect
 
+
+""""
+    fmap_with_path(f, x, ys...; exclude = isleaf, walk = DefaultWalkWithPath(), [prune])
+
+Like [`fmap`](@ref), but also passes a `KeyPath` to `f` for each node in the
```
darsnack, line 316:
```suggestion
Like [`fmap`](@ref), but also passes a [`KeyPath`](@ref Functors.KeyPath) to `f` for each node in the
```

src/Functors.jl
```diff
@@ -303,4 +309,46 @@ julia> fcollect(m, exclude = v -> Functors.isleaf(v))
...
+recursion. The `KeyPath` is a tuple of the indices used to reach the current
+node from the root of the recursion. The `KeyPath` is constructed by the
+`walk` function, and can be used to reconstruct the path to the current node
+from the root of the recursion.
+
+`f` has to accept two arguments: the value of the current node and the associated `KeyPath`.
```
darsnack, line 322:
```suggestion
`f` has to accept two arguments: the associated `KeyPath` and the value of the current node.
```

src/maps.jl
```diff
@@ -11,7 +11,22 @@ function fmap(f, x, ys...; exclude = isleaf,
   execute(_walk, x, ys...)
 end
 
-fmapstructure(f, x; kwargs...) = fmap(f, x; walk = StructuralWalk(), kwargs...)
+function fmap_with_path(f, x, ys...; 
+                exclude = isleaf,
+                walk = DefaultWalkWithPath(),
+                cache = IdDict(),
+                prune = NoKeyword())
```
darsnack, line 18: Nitpick: tab alignment

darsnack (approved):
Fair but @ToucheSir may want to weigh in on these changes and re: breakage now or later.

ToucheSir:
My only comment for now would be whether separate `_with_path` functions were necessary, or if we could've gotten away with something like passing the keypath in as a keyword argument. We should really look into API cleanup because the increase in API surface area and code duplication is noticeable now. That however is a different issue/PR discussion.

PR #78, opened 2024-03-11
reorganize API docs

CarloLucibello:
Introducing an index and breaking down the docstrings into different sections.


ToucheSir (approved):
An index is always nice. One chore that this brings up is that it'd be nice to have docs previews enabled. Will open an issue for that.

PR #80, opened 2024-03-13
functor `Base.Splat`

mcabbott:
Context is revisiting all kinds of multiple in/out vs tuple things in Flux in https://github.com/FluxML/Flux.jl/pull/2393

### DIFF
--- a/src/base.jl
+++ b/src/base.jl
@@ -9,7 +9,10 @@
 @functor Base.Fix1
 @functor Base.Fix2
 @functor Base.Broadcast.BroadcastFunction
-@functor Base.Splat
+
+@static if VERSION >= v"1.9"
+  @functor Base.Splat
+end
 
 @static if VERSION >= v"1.7"
   @functor Base.Returns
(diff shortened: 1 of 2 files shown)


src/base.jl
```diff
@@ -9,6 +9,7 @@
 @functor Base.Fix1
 @functor Base.Fix2
 @functor Base.Broadcast.BroadcastFunction
+@functor Base.Splat
```
mcabbott, line 12:
```suggestion

@static if VERSION >= v"1.9"
  @functor Base.Splat
end
```

test/base.jl
```diff
@@ -56,6 +56,12 @@ VERSION >= v"1.7" && @testset "Returns" begin
   @test Functors.functor(ret)[2]((value = 1:3,)) === Returns(1:3)
 end
 
+@testset "Splat" begin
```
mcabbott, line 59:
```suggestion
VERSION >= v"1.9" && @testset "Splat" begin
```

CarloLucibello (approved):

PR #82, opened 2024-05-09
Improve type stability of cached walks

chengchingwen:
This PR adds a special cache type that allows the compiler to use the signature of the un-cached `walk` to generate corresponding type assertion to the untyped cache (`IdDict{Any, Any}`). This would improve the type stability of `fmap`and friends. It also looses the constraint of the cache type so functionality outside `fmap` remains the same.

### DIFF
--- a/src/Functors.jl
+++ b/src/Functors.jl
@@ -7,6 +7,7 @@ export @functor, @flexiblefunctor, fmap, fmapstructure, fcollect, execute, fleav
 include("functor.jl")
 include("keypath.jl")
 include("walks.jl")
+include("cache.jl")
 include("maps.jl")
 include("base.jl")
 
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -11,23 +11,48 @@ Base.iterate(cache::WalkCache, state...) = iterate(cache.cache, state...)
 Base.setindex!(cache::WalkCache, value, key) = setindex!(cache.cache, value, key)
 Base.getindex(cache::WalkCache, x) = cache.cache[x]
 
-function __cacheget_generator__(world, source, self, cache, x, args #= for `return_type` only =#)
-  # :(return cache.cache[x]::(return_type(cache.walk, typeof(args))))
-  walk = cache.parameters[3]
-  RT = Core.Compiler.return_type(Tuple{walk, args...}, world)
-  body = Expr(:call, GlobalRef(Base, :getindex), Expr(:., :cache, QuoteNode(:cache)), :x)
-  if RT != Any
-    body = Expr(:(::), body, RT)
-  end
-  expr = Expr(:lambda, [Symbol("#self#"), :cache, :x, :args],
-              Expr(Symbol("scope-block"), Expr(:block, Expr(:meta, :inline), Expr(:return, body))))
-  ci = ccall(:jl_expand, Any, (Any, Any), expr, @__MODULE__)
-  ci.inlineable = true
-  return ci
-end
-@eval function cacheget(cache::WalkCache, x, args...)
-    $(Expr(:meta, :generated, __cacheget_generator__))
-    $(Expr(:meta, :generated_only))
+@static if VERSION >= v"1.10.0-DEV.609"
+    function __cacheget_generator__(world, source, self, cache, x, args #= for `return_type` only =#)
+        # :(return cache.cache[x]::(return_type(cache.walk, typeof(args))))
+        walk = cache.parameters[3]
+        RT = Core.Compiler.return_type(Tuple{walk, args...}, world)
+        body = Expr(:call, GlobalRef(Base, :getindex), Expr(:., :cache, QuoteNode(:cache)), :x)
+        if RT != Any
+            body = Expr(:(::), body, RT)
+        end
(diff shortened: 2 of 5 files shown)


CarloLucibello:
This adds some complexity to the code and some fragility as well, since it seems it could break with newer julia versions.
Can you post some benchmarks showing performance improvements?

chengchingwen:
Not a benchmark, but without this PR:
```julia
julia> @code_warntype gpu(Chain(Dense(3, 5), Dense(5, 2)))
MethodInstance for Flux.gpu(::Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}})
  from gpu(x) @ Flux ~/.julia/packages/Flux/Wz6D4/src/functor.jl:248
Arguments
  #self#::Core.Const(Flux.gpu)
  x::Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}
Body::Chain{T} where T<:Tuple{Any, Any}
1 ─ %1 = Flux.FluxCUDAAdaptor()::Core.Const(Flux.FluxCUDAAdaptor(nothing))
│   %2 = Flux.gpu(%1, x)::Chain{T} where T<:Tuple{Any, Any}
└──      return %2
```
v.s. with:
```julia
julia> @code_warntype gpu(Chain(Dense(3, 5), Dense(5, 2)))
MethodInstance for Flux.gpu(::Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity
), Matrix{Float32}, Vector{Float32}}}})
  from gpu(x) @ Flux ~/.julia/packages/Flux/Wz6D4/src/functor.jl:248
Arguments
  #self#::Core.Const(Flux.gpu)
  x::Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vecto
r{Float32}}}}
Body::Union{Chain{Tuple{Dense{typeof(identity), CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, CuArray{Float32, 1, CUDA.Mem.D
eviceBuffer}}, Dense{typeof(identity), CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, CuArray{Float32, 1, CUDA.Mem.DeviceBuff
er}}}}, Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}}
1 ─ %1 = Flux.FluxCUDAAdaptor()::Core.Const(Flux.FluxCUDAAdaptor(nothing))
│   %2 = Flux.gpu(%1, x)::Union{Chain{Tuple{Dense{typeof(identity), CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, CuArray{Fl
oat32, 1, CUDA.Mem.DeviceBuffer}}, Dense{typeof(identity), CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, CuArray{Float32, 1,
 CUDA.Mem.DeviceBuffer}}}}, Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}}
└──      return %2
```

CarloLucibello:
@darsnack @ToucheSir what do you think? I'm unfamiliar with expression manipulations.

darsnack:
I am also concerned about fragility. The implementation itself is sensible, but as written seems like it will need to get updated for internal changes often. The core idea is to use the return type of the walk to force the type when accessing the cache, right? That seems like a very straight-forward generated function to write with the call to `return_type` being the only brittle bit. Or is the rest of the current implementation necessary for performance reasons? Accessing the `IdDict` is the main reason Functors is type unstable, so fixing it is nice.

Pulling back, is there a use-case where we lack a function barrier between the call to `gpu` and the hot code path?

chengchingwen:
>  The core idea is to use the return type of the walk to force the type when accessing the cache, right? That seems like a very straight-forward generated function to write with the call to return_type being the only brittle bit. Or is the rest of the current implementation necessary for performance reasons?

Yes, essentially the whole generated function is just to generate `return cache.cache[x]::(return_type(cache.walk, typeof(args)))`. It also seems to be doable without generated function, but with the generated function we can get the precise world-age (though I'm not familiar enough with the world-age mechanism to know if the precise world-age is required in this use-case). 

> Pulling back, is there a use-case where we lack a function barrier between the call to gpu and the hot code path?

if you need to handle data movement during the forward/backward pass.

CarloLucibello:
given the concerns expressed in https://github.com/LuxDL/Lux.jl/issues/1017 I think we should do this.

CarloLucibello (approved):

chengchingwen:
@CarloLucibello Since Julia v1.10 is the new LTS, do you think we could drop v1.6 support so that we can remove that `@static if VERSION >= v"1.10.0-DEV.609"` branch which makes the code look fragile?

CarloLucibello:
yes, we should do that.

PR #83, opened 2024-05-13
add setkeypath!

CarloLucibello:


darsnack:
I assume this is to set tree leaves to zero? Why not `fmap!` that applies a function to leaves without reconstructing?

CarloLucibello:
`setkeypath!` is more flexible than `fmap!`, e.g. it allows changing scalar leaves or replacing entire subtrees.

CarloLucibello:
merge?

darsnack (approved):

PR #100, opened 2026-02-18
AbstractSet is functor

kleinschmidt:
Currently, `Base.Set` and possibly other AbstractSets are treated as normal
structs.  Since `Base.Set{K}` wraps a `Dict{K,Nothing}`, the Set constructor
treats its input as an iterable of values, and `Dict` iterates `Pair`s, `Set`s
end up as `Set{Pair{T,Nothing}}` after walking.

This adds tests to ensure that `Set` and `OrderedSet` (on behalf of all
`AbstractSet`s everywhere) survive `fmap`.

I'm not 100% sure this is the right way to handle these (generally) un-ordered,
not-natively-`map`able collections (I think you might get surprising results
using `fmap` with multiple arguments for instance.  I can work aroudn this with
some lite type piracy but wanted to open this up for discussion anyway!


CarloLucibello:
seems reasonable

PR #104, opened 2026-08-05
Use Meta.lower instead of ccall(:jl_expand)

IanButterworth:
Claude:

---

Julia 1.13 moved lowering behind `Core._lower` and the `jl_expand` symbol no longer exists, so `__cacheget_generator__` failed with "could not load symbol jl_expand" on every `fmap` call.

`Meta.lower(m, x)` is the public wrapper for the same operation: it calls `ccall(:jl_expand, Any, (Any, Any), x, m)` up to 1.12 and `Core._lower` from 1.13. It is therefore used unconditionally on all supported versions rather than behind a VERSION branch, and is an exact equivalent of the previous ccall on 1.10-1.12.


simeonschaub (approved):

simeonschaub:
Do you also want to bump the version in the Project.toml, so we can get a new release out right away?

CarloLucibello:
version bumbed in a pr I just merged, so I'll merge also this and tag a release

IanButterworth:
Thanks. Would you mind reporting here when the release is tagged, as this unblocks some work. Much appreciated.

CarloLucibello:
done!

